PR #2, opened 2018-02-18
Add Google Assistant

HarrisJT:
Couldn't find an image, but still worth listing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@ Feel free to send a PR! :smile:
 
 - [Alexa](#alexa)
 - [Codeship](#codeship)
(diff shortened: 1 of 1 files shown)


swapagarwal:
Thanks @HarrisJT! 👍

swapagarwal:
I removed the $200 credits part as it isn't a tangible form of swag. What do you think?

swapagarwal:
Continued in #3

HarrisJT:
I wasn't sure so I put it in parenthesis, up to you. I would take 200$ of US Government swag ;)

PR #7, opened 2018-02-19
Seperate Active and Inactive Promotions

kevin51jiang:
For people who want to filter out inactive promotions.

Hacktoberfest website says "Check back next October for details on how to participate!" so I marked it as inactive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,12 +5,14 @@ As a developer, nothing piques my interest as much as developer swag! This is an
 Feel free to send a PR! :smile:
 
 ## Contents
-
+###Currently Active
 - [Alexa](#alexa)
 - [Codeship](#codeship)
 - [Google Assistant](#google-assistant)
-- [Hacktoberfest](#hacktoberfest)
 - [npm](#npm)
+
+###Inactive
+- [Hacktoberfest](#hacktoberfest)
 
 ## Alexa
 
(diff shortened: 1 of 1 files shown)


jgierer12:
Perhaps those could be wrapped in a `<details>` tag or put into a separate file to clean up the readme a bit. I imagine it will get quite cluttered once more things are added.

swapagarwal:
Maybe we can add a badge similar to Digital Ocean?

@kevin51jiang @jgierer12 @pkalemba What do you think? 😄

d4nyll:
My $0.02 - most of the time, visitors to the page wants to know opportunities they can enter _now_; having separated sections is going to make the page easier to scan through than having a badge.

kevin51jiang:
This might seem a bit too complicated, but could we turn promos into seperate Json or text files or something, and make a directory to put them in? These files would have spots for tags, image url, status etc. Then get an app to run daily that will construct 3 different .md files. One for all, one in progress, and one for ended promos.

rbstrachan:
I find this slightly pointless given that `Oktoberfest` is the only link that would be separated. This feature would come in handy if the list of inactive promotions was substantial so that the active promotions could be seen without constant interruption.
On top of that, it is made clear both in the title and the description that the promotion only runs in October, so labelling it `Inactive` also seems pointless.
Considerably changing the layout of the README for one link may actually make it more obtrusive and unintuitive to the user, rather than helping them read it, however, this is just my opinion.

antonio6643:
Perhaps if more time-based promotions are added, we could add an "Upcoming Promotions" section.

swapagarwal:
All promotions are currently active! 😇

PR #19, opened 2018-02-20
Removing digital ocean

juancri:
The survey has ended

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,6 @@ Feel free to send a PR! :smile:
 - [Alexa](#alexa)
 - [Codeship](#codeship)
 - [devRant](#devrant)
-- [Digital Ocean](#digital-ocean)
 - [Google Assistant](#google-assistant)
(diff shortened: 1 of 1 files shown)


swapagarwal:
I was thinking if we should just move it to a separate section (expired) below instead of removing it completely. This would help in keeping a tab on the link when it becomes active again! What do you think?

swapagarwal:
How about adding something like this?

karanrajpal14:
@swapagarwal makes sense. Go for it! :+1:

swapagarwal:
Thanks @juancri and @karanrajpal14! 👍

vikaspotluri123:
You might want to add it to the ToC as well 🤔

swapagarwal:
@vikaspotluri123 I added there first, but faced some problems in linking to the corresponding section. Working into it..

swapagarwal:
Done! 🎉

PR #27, opened 2018-02-22
use table to display status of all opportunities

butlerx:
try to make it clearer the status of all the opportunities by adding badges to them all and a table

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,51 +8,53 @@ Feel free to send a PR! :smile:
 
 ## Contents
 
-- [Alexa](#alexa)
-- [Codeship](#codeship)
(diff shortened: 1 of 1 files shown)


swapagarwal:
Hmm, a table makes things clunky. I'd rather vote for keeping it simple. See @rbstrachan [comment](https://github.com/swapagarwal/swag-for-dev/pull/7#issuecomment-367510306).
The only reason DO and Hacktoberfest is on the list is because they will be back later this year, otherwise I'd have removed it entirely. So ideally, all promotions on the list should be active only.
What do others think?

swapagarwal:
All promotions are currently active! 😇

PR #38, opened 2018-04-07
Update Alexa section

zac-garby:
It looks like the terms have changed. From the reference: "If your skill attracts 1,000 unique customers in its first 30 days in the Alexa Skills Store, you will also earn an Echo Spot to help you make Alexa even smarter."


swapagarwal:
Thanks @zac-garby! 👍

PR #47, opened 2018-07-10
devRant changed amount of ++'s

MrWouterNL:
- [X] I've checked that there isn't a duplicate pull request.

devRant now requires 750++'s for a stressball and 30++'s for a stickerpack. Reference: https://devrant.com/free-stickers


swapagarwal:
Thanks @MrWouterNL! 👍

PR #51, opened 2018-09-10
Update README.md

jpita:
according to https://hacktoberfest.digitalocean.com/ it is five and not four

- [x] I've checked that there isn't a duplicate pull request.

Uncomment the corresponding section below.


swapagarwal:
Thanks @jpita! 👍
Last year, it was four.. 😇

PR #53, opened 2018-09-29
Removing digital ocean as their survey has ended

oliviernt:

**Something else**

Removing digital ocean from data.json as their survey has ended


swapagarwal:
Thanks @oliviernt! 👍

Issue #52, opened 2018-09-20
Design the website


swapagarwal:
We shall be hosting this on either `devswag.xyz` or `devswag.io` domain.

(Moving this from #43 to its separate thread)
What extra features would you like to see on the website?

One example I can think of is the ability to filter swag opportunities by the level of difficulty. Feel free to share more!

Also, it'd be great if someone volunteers to get the initial design of the website in place. Cheers!

zac-garby:
I’d like to design the website :)

cozyplanes:
`io` seems easier to remember

swapagarwal:
@zac-garby Awesome! How about you take the lead on this? Let's keep it minimalistic and get it out before Hacktoberfest. 😇

zac-garby:
Cool! I'll start it a bit later today when I get home. Do we want it to automatically fetch the swag opportunities from the repo?

swapagarwal:
@zac-garby I can export the list in json format if that helps?
Something like:
```
{
    "name": "Alexa",
    "difficulty": "hard",
    "description": "Publish a skill, get a hoodie (varies by country) and an Echo Spot (if 1,000 unique customers use your skill in its first 30 days of publication)!",
    "reference": "https://developer.amazon.com/alexa-skills-kit/alexa-developer-skill-promotion",
    "image": "https://images-na.ssl-images-amazon.com/images/G/01/mobile-apps/dex/alexa/alexa-skills-kit/ASM0761-Dev-LP-Goodiepage-UK-March._CB501585903_.png"
}
```

zac-garby:
@swapagarwal Yeah, that would be very useful - thanks

swapagarwal:
Your wish, is my command: https://github.com/swapagarwal/swag-for-dev/blob/master/data.json
The website can load data from https://raw.githubusercontent.com/swapagarwal/swag-for-dev/master/data.json

plibither8:
Hey @swapagarwal @zac-garby, if possible, I'd like to help too :)

swapagarwal:
@plibither8 Sure! It'd be great if you can sketch up a quick mock and share the design that's on your mind.

plibither8:
Sure, I'll start work late at night. I'll try to share it by tonight (IST) itself

zac-garby:
@swapagarwal I've made a quick design of what the site might look like. I was thinking for the colour scheme we could keep it really simple, maybe just black and white.

[...]

PR #54, opened 2018-09-30
Website for swag-for-dev

plibither8:
Conversation regarding the website's development: #52


plibither8:
I think deployment to Netlify is failing because it is linked with Jekyll. We might have to fix that @swapagarwal

swapagarwal:
`LGTM`
Let's add those css changes to fit images as seen in the screenshot and we'll make it live!

plibither8:
@swapagarwal Done!

swapagarwal:
Quick Question: Shouldn't `site` folder be in .gitignore?

plibither8:
The site is being deployed from the `dist` folder, so it has to be included in the repo

swapagarwal:
So basically, Netlify provides an option to specify a `build` command which it will run after every commit. The output should not reside the repository itself as it is auto-generated (i.e. we won't make any changes to it manually).

plibither8:
Ah alright, makes sense. So what will the build command be?

swapagarwal:
I think you'll have to add a script in package.json which runs your 3 gulp tasks.
Try deleting dist folder, and it should be generated when you run `npm run build`

swapagarwal:
I've set the build command in netlify to `cd site && npm install && npm run build`

plibither8:
It's working 👍

PR #59, opened 2018-10-01
Download and optimize swag images on build

### LINKED ISSUE #57: Optimize images
Images take too long to load on slower networks!
They can be compressed and stored on the server itself, instead of loading it every time from different servers.

aslafy-z:
Closes #57 

Implements 4 new gulp tasks:
- `swag-img:download`: Download all images and minify them
- `swag-img:update-data`: Copy data.json file to assets and change path to local images
- `swag-img`: Run the two subtasks


swapagarwal:
@aslafy-z You might want to move `data.json` to `site` directory.

aslafy-z:
Actually I prefer not to, it's like the repo's source of trust.  It is not necessary in any case

aslafy-z:
https://deploy-preview-59--peaceful-chandrasekhar-efde8e.netlify.com/assets/js/index.js does not exists, I don't know why

plibither8:
That's an odd issue. Maybe force re-deploy (if we can)?

aslafy-z:
I just did it, it builds and deploys: https://app.netlify.com/sites/peaceful-chandrasekhar-efde8e/deploys/5bb25ffe73f2cf7ff2060e69, but it looks like it's missing some files

aslafy-z:
Fixed it, looks good to me

aslafy-z:
we could add image-resize later

swapagarwal:
Nice!
The total size of images is still well over 1 MB. Let's explore ways to optimize it.

site/gulpfile.js
```diff
@@ -80,9 +81,10 @@ gulp.task('swag-img:update-data', () => {
         s.image = `/assets/swag-img/${name}`;
         return s;
     });
-    const payload = JSON.stringify(newSwagList);
-    fs.writeFileSync('dist/assets/data.json', payload);
     SWAG_LIST = newSwagList;
+    const stream = source('data.json');
```
vikaspotluri123: Is there a reason to use streams here in place of fs? Since Gulp supports callbacks (and promises as well, but using callbacks here for simplicity)

```js
const {writeFile} = require('fs');

// ...
gulp.task(/*...*/, cb =>
/*...*/
return writeFile('./dist/assets/data.json', JSON.stringify(newSwagList), cb);
```

aslafy-z: I used streams to be sure that the folder is being created by gulp. But I can revert the commit and use asynchronous functions instead, sure.

aslafy-z: I agree that adding a dependency for this unique thing may not be necessary

aslafy-z: Do you think this is better ?
```js
return mkdir('dist/assets', (err) => {
	// if (err) throw err;
	writeFile('dist/assets/data.json', JSON.stringify(newSwagList), cb)
});
```

site/gulpfile.js
```diff
@@ -1,17 +1,26 @@
...
+const webserver     = require('gulp-webserver');
+const concat        = require('gulp-concat');
+const download      = require('gulp-download-stream');
+const imagemin      = require('gulp-imagemin');
+const source        = require('vinyl-source-stream');
+const buffer 				= require('vinyl-buffer');
+
+const filenameSafe  = s => s.replace(/[^a-z0-9.]/gi, '_').replace(/_{2,}/g, '_').toLowerCase();
+
+let SWAG_LIST       = JSON.parse(fs.readFileSync('../data.json'));
```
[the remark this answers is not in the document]
aslafy-z: Yep, I changed that

site/src/js/index.js
```diff
@@ -46,10 +46,10 @@ const renderSwag = swag => {
         })
         .sort((a, b) => {
             switch (sorting) {
-                case 'Alphabetical':            return a.name.toLowerCase() > b.name.toLowerCase();
-                case 'Alphabetical, reversed':  return a.name.toLowerCase() < b.name.toLowerCase();
-                case 'Difficulty':              return difficultyIndex(a.difficulty) > difficultyIndex(b.difficulty);
-                case 'Difficulty, reversed':    return difficultyIndex(a.difficulty) < difficultyIndex(b.difficulty);
+            case 'Alphabetical':            return a.name.toLowerCase() > b.name.toLowerCase();
```
RealOrangeOne: would suggest not reformatting like this, as it seems unnecessary to the PR

plibither8: I'm guessing the linter is doing that.

RealOrangeOne: Looking at the whitespace removal, i think you're right. Still shouldn't commit unnecessary changes like this, unless these changes were causing linter errors, that is.

site/gulpfile.js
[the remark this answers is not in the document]
vikaspotluri123: Is the commit live? I don't see it 🙈

site/src/js/index.js
swapagarwal: @plibither8 The linter (which you had put) is universal, right? Do you know why it is behaving differently for him?

plibither8: I had forgot to run the linter before commiting the file :sweat_smile:

swapagarwal: Haha, np.

aslafy-z: Yes, i'm using atom with `linter-eslint` on save. It follows the local `.eslintrc` file so it should respect the rules.

https://eslint.org/docs/rules/indent

Here we can see : `"SwitchCase" (default: 0) enforces indentation level for case clauses in switch statements`
And: `Indent of 2 spaces with SwitchCase set to 0 will not indent case clauses with respect to switch statements.`

Althrough I see Ident configured on 4 spaces in the `.eslintrc`, I dont know why it's not indenting the `case` statements.

site/gulpfile.js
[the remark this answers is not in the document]
aslafy-z: It's live now

aslafy-z:
220 KB now, I tried to crop but depending on images format it was not fitting well.
Limiting the height to 300px seems good.

swapagarwal:
Wow, that's great @aslafy-z!!

plibither8:
I'm pretty sure with this, #64 becomes invalid :smile: 

Though instead of the placeholder-loading blocks, we can have "Loading..." as a text. This way we won't have to wait for the CSS file of (zalog/placeholder-loading)[https://github.com/zalog/placeholder-loading] to load (which would be an additional HTTP request).

kevin51jiang:
There's a very noticeable quality difference on bigger screens. Is there any way so that we limit the optimizations to mobile devices?

aslafy-z:
Sure, we could generate different images for both sizes with
gulp-responsive and use  and srcset to select the one we show.

Le mar. 2 oct. 2018 à 20:41, Kevin Jiang <notifications@github.com> a
écrit :

swapagarwal:
Hey @kevin51jiang, thanks for chiming in! Can you post a screenshot?

site/gulpfile.js
```diff
@@ -20,42 +28,100 @@ gulp.task('pug', () => {
...
         .pipe(gulp.dest('dist/assets/css'));
-})
+});
 
 gulp.task('js', () => {
     return gulp.src('src/js/*.js')
         .pipe(concat('index.js'))
         .pipe(babel({
-            "presets": [
+            'presets': [
```
vikaspotluri123: Since this is an object quotes aren't required at all

aslafy-z: Fixed

swapagarwal:
@vikaspotluri123 I was waiting to review this as @kevin51jiang had raised a concern regarding image quality. 😅

kevin51jiang:
Woops sorry for not responding. It mostly affects bigger monitors which a screenshot would not be able to illustrate

aslafy-z:
Switching from `80%` to `90%` jpeg quality increase `swag-img` folder from `220KB' to `324KB`.

swapagarwal:
Pinging @kevin51jiang 👋

aslafy-z:
I think we can merge this, I switched to 90% quality and it's way better than before.

swapagarwal:
Thanks a lot @aslafy-z!
This is really a major optimization. 🎉

Issue #75, opened 2018-10-03
New Page Idea: devswag.io/hacktoberfest


swapagarwal:
There are quite a lot of interesting hacktoberfest swag opportunities this year! I was thinking of having a quick way to access them.
What do you guys think of having a new page `/hacktoberfest` specifically for this purpose?

aslafy-z:
I think it would be great. We will have to add labels to the swags to do so, right ?

Jacob8765:
Would they still be visible on the main page, or would they only be accessible from the hacktoberfest page?

plibither8:
It's a great idea, but, how are we going to _advertise_ it? Since this is a limited-time page, would the amount a traffic visiting that /hacktoberfest vs traffic visiting the main page be worth it? Just wondering

zac-garby:
> Would they still be visible on the main page, or would they only be accessible from the hacktoberfest page?

They should definitely be visible on both pages.

swapagarwal:
@aslafy-z Yes, we would need to add labels for this.
@Jacob8765 @zac-garby Everything would be visible on main page; filtered swag on hacktoberfest page for quick access. (Same can be achieved by adding a `show only hacktoberfest opportunities` checkbox on the main page but having a new page makes sharing easier.)
@plibither8 A button can be shown on main page linking to the hacktoberfest page. (Just thinking out loud here)

plibither8:
> Yes, we would need to add labels for this.

I'm guessing labels in the data.json file, right?

Jacob8765:
Maybe instead of adding a new page, it would be easier to use a URL query - something like devswag.io?filter=hacktoberfest.

darshkpatel:
IMHO We can just have a hacktoberfest filter on the main page.

Also can have something like devswag.io/filter=hacktoberfest , in case we
need a sharable link

swapagarwal:
@Jacob8765 @darshkpatel Makes sense.

plibither8:
Should I start work on this?

aslafy-z:
@plibither8 I think it would be better to add something like 
```diff
labels: ['hacktoberfest'],
```
To the data.json file as we could have other categories in the future

plibither8:
That's a good idea, I'll initialise all swag items with an empty `labels` array, except the hacktoberfest ones

swapagarwal:
Sure, go ahead!
How about using `tags` instead of `labels`?

Jacob8765:
[...]

PR #78, opened 2018-10-03
feat: Implement tag handling

plibither8:

Fill the corresponding section below:

(Something else)  

As discussed in #75 this PR implements a checkbox on the homepage beneath the sort and filter options to display "only hacktoberfest opportunities". Additionally, [devswag.io/?filter=hacktoberfest](https://devswag.io/?filter=hacktoberfest) will check the checkbox and show hacktoberfest swag items.


plibither8:
Just a thought: to stay consistent with our data.json structure, I suggest having the URL as devswag.io/?**tag**=hacktoberfest rather than devswag.io/?**filter**=hacktoberfest

swapagarwal:
When I click on the checkbox, no items are visible.

plibither8:
The PR uses an updated data.json, but the code is fetching the data.json which in the `master` branch. Once it's merged, you'll be able to see items.

swapagarwal:
When you click the checkbox, the URL should update accordingly.
Similarly, when you visit the URL, the corresponding fields should be checked.

I was thinking of making this generic enough so that we only need to add more tags in the json file in future w/o any code changes. (something like fetch all tags from json file and show checkbox next to each: filter opportunites: -checkbox- hacktoberfest -checkbox- abc -checkbox- xyz, etc.)

swapagarwal:
It would be on both:
When you visit a URL, tags need to be prepopulated from the query params.
When you update (select / unselect) tags, query string should be modified accordingly (for easy sharing!).

swapagarwal:
We can also merge swagType with this, so you'll checkboxes of t-shirts, hacktoberfest, stickers, etc.
What do you think?

plibither8:
About making it generic, I was thinking the same too. So this is what I propose: 

1. Move the items in `swagType` in the data.json file to `tags` and remove `swagType`. This removes the two classificiation types we have thereby causing less confusion.

2. Tags can be displayed under the title text of each swag item like that displayed when adding GitHub labels.

3. A "Show only" dropdown can be implemented (or a checklist) that will give the option to sort by tags (hacktoberfest|stickers|ccessories|clothing|other). Here, we can aim to make the tags more generic, like instead of "Echo Spot" as a swagtype, we can have "devices" or something more befitting.

swapagarwal:
1. 👍 
1. 👍
1. Checklist seems more appropriate here, as I may want to filter out hacktoberfest along with stickers. Sorting is already being taken care of (name / difficulty). We just need to add filters based on tags.

Haha, no issues. Try using a GitHub client like FastHub.

plibither8:
Sure! I'll continue work with this when I get back home, gotta go to school now :|

Yeah, I'll use FastHub now, hah

aslafy-z:
Maybe instead of having a checkbox per swagtype which can eavily bloat the interface, we could use something like this : 
http://wenzhixin.net.cn/p/multiple-select/docs/ (without the jQuery deps, sure). It would allow users to select multiple swag type without adding too much bloat to the ui

plibither8:
> Edit: https://github.com/Mobius1/Selectr looks good for me, 6.5kB minified & gzip, WDYT ?

Looks good! We can surely afford that.

swapagarwal:
@aslafy-z +1 from my side as well!
P.S. Just out of curiosity, is there a go-to place for searching these types of replacements? 😜

aslafy-z:
I usually uses Google + GitHub, it's the best place to search but need to get the good keywords. For software alternatives I uses alternativeto. For this special case (jquery-free alternatives), I heard about http://youmightnotneedjquery.com/ but I didn't used it to find Selectr.

plibither8:
@swapagarwal @aslafy-z I have implemented the tag handling as requested, please go over the changes.

swapagarwal:
In the preview url, the request to `data.json` is failing with 404.

site/src/pug/includes/head.pug
```diff
@@ -14,4 +14,7 @@ link(rel="apple-touch-icon" size="128x128" href="/assets/img/logo.png")
 link(rel="stylesheet" href="/assets/css/index.css")
 link(rel="stylesheet" href="https://fonts.googleapis.com/css?family=Cabin")
 
+link(href='https://unpkg.com/mobius1-selectr@latest/dist/selectr.min.css' rel='stylesheet' type='text/css')
+script(src='https://unpkg.com/mobius1-selectr@latest/dist/selectr.min.js' type='text/javascript')
```
vikaspotluri123: Although it probably won't be an issue you should probably lock the version

plibither8:
@swapagarwal just realised the I forgot to change `DATA_URL` back to the GitHub link instead of the development link I was using. I'll push the change soon.

site/src/pug/includes/head.pug
plibither8: Good point!

vikaspotluri123: To extend this, it's probably also worth adding [SRI](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) to external [css](https://www.w3.org/TR/SRI/#resource-integrity) and js

site/src/pug/includes/body.pug
```diff
@@ -11,23 +11,25 @@ include ./fork.pug
 
         .buttons.flex
 
-            .item.flex
-                #codefund_ad
```
swapagarwal: This is needed to cover up our hosting costs. 😅

plibither8: Whoops! Didn't realise I deleted that :sweat_smile:

swapagarwal:
Thanks Zadkiel and Vikas for reviewing.
Special Thanks to Mihir for taking this idea to production! 🤖

site/src/pug/includes/body.pug
swapagarwal: No problem; check out the new layout (bottom bar). Let me know if you think top bar is a more suitable place..

plibither8: The ad goes out-of-screen currently because it is nested in the `.buttons` div. Also, the bottom part of the screen is getting covered by the ad, so some margin will be required.

👍 for bottom bar though!

swapagarwal: So I should probably move it out to the top-most level, right? Not sure how to add margin though!

plibither8: Should I create a PR? It's a minor fix

swapagarwal: I think I managed to fix it. Can you check?

plibither8:
#83 is dealing with that I guess.

aslafy-z:
Oops, I thought it was already merged. Sorry.

site/src/pug/includes/body.pug
swapagarwal: Oh, I thought you wanted to add margin between banner and site content 😂

Issue #64, opened 2018-10-02
Display loading gif while content is being loaded on the website


plibither8:
- [x] I've checked that this isn't a duplicate issue.

On (relatively) slower internet connections, there is upto a 5-7 second delay from when the website frontend (title text and footer) has loaded to when the content (the JSON data and images) has finally loaded. I suggest we display a 'loading' GIF until the files have been downloaded (using event listeners) so that the page doesn't look blank, and the user waits.

aslafy-z:
Maybe we could use some placeholder loading animation to preallocate some space for the swag items. It could be made with https://github.com/zalog/placeholder-loading

plibither8:
That looks nice! Mind if I take this issue?

aslafy-z:
Sure, go ahead !

plibither8:
@aslafy-z @swapagarwal Is this enhancement still required? @aslafy-z did a great job with #59, which I believe should not cause the problem for which this issue was originally created.

aslafy-z:
I think this is still needed as it can take some times to load the data.json file depending on connection. We should at least add a text placeholder Loading... while loading the data.json.
Also, I tried to integrate https://afarkas.github.io/lazysizes/ but I'm not sure we need it.

plibither8:
> Maybe we could server-render a part of the data.json list

Oh boy, it would be great if we could get this happening... on netlify. Two advantages I can thinking of off the top of my head:

1. Shorter load time and less lag
2. Better for SEO as the actual swag items and the content will be allowed to be crawled by search engines.

aslafy-z:
It could be done on website deployment, as adding swag triggers netlify deployment right ?

swapagarwal:
If the gulp pipeline can build the static site, that would be perfect! So basically the layout of the items / cards along with the text / description can be prerendered and only images need to be loaded on page load.

swapagarwal:
@plibither8 Let me know if you think we still need this after image optimization and SSR. 😇

vikaspotluri123:
@plibither8 I think we can close this? With all of the improvements we've made over the last month or so, I don't think a loading indicator is necessary anymore 🎉

plibither8:
Supremely sorry for the delay :neutral_face:! Certainly, this is not needed anymore. I'll close it.

PR #84, opened 2018-10-04
✨ server side rendering

vikaspotluri123:

 - Prerender swag to prevent content flashing (refs #64, may or may not close it)
 - Load swag as JSON at bottom of file (removes need for additional HTTP request)
	- This is really tiny (2.2k) so it should have a minimal impact on load times

Todo (not in this PR)

 - [ ] filter: toggle visibility based on classes to reduce re-rendering requirement
 - [ ] sort: Clone DOM nodes rather than use template literals. This is to DRY up and possibly get performance improvements
 - [ ] Disallow sorting by difficulty when only one difficulty is shown

### DIFF
--- a/site/src/pug/includes/body.pug
+++ b/site/src/pug/includes/body.pug
@@ -32,6 +32,20 @@ include ./fork.pug
                     select#tags
 
         .content.flex#content
+            - let difficulty;
+            each swag in swagList
+                - difficulty = swag.difficulty + ' difficulty';
+                div.item
+                    div.title.flex
+                        h1=swag.name
+                        div(class=difficulty title=difficulty)
+                    p.swag
+                        each swagTag in swag.tags
+                            span!=swagTag
+                    div.flex.img-container
+                        img(src=swag.image)
+                    p.description!=swag.description
+                    a(href=swag.reference) Check it out
 
     footer.flex
         .item.flex
@@ -42,5 +56,8 @@ include ./fork.pug
             p Designed by #[a(href="https://github.com/zac-garby") Zac Garby] and #[a(href="https://mihir.ch/") Mihir Chaturvedi]
         .item.flex
             p Want to be listed? #[a(href="https://github.com/swapagarwal/swag-for-dev/issues/new") Contact us].
-
     #codefund_ad
+
+    script(defer).
+        window.swag = !{JSON.stringify(swagList)};
+        window.swagTags = !{JSON.stringify(tags)};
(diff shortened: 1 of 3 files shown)


site/src/pug/includes/body.pug
```diff
@@ -30,6 +30,18 @@ include ./fork.pug
...
+            each swag in data
+                - difficulty = swag.difficulty + 'difficulty';
+                div.item
+                    div.title.flex
+                        h1=swag.name
+                        div(class=difficulty title=difficulty)
+                    p.swag=swag.swagType
+                    div.flex.img-container
+                        img(src=swag.image)
+                    p.desciption=swag.description
```
aslafy-z: should be `p.desciption!=swag.description`

aslafy-z:
This also needs to wait for #59

site/src/pug/includes/body.pug
vikaspotluri123: Thanks, fixed!

vikaspotluri123:
I'll rebase sometime in the next 12 hrs hopefully

vikaspotluri123:
That was... interesting! Rebase is complete 😄

vikaspotluri123:
@swapagarwal I think this is ready to be merged 🎉

vikaspotluri123:
I implemented it in a different method

Zadkiel was suggesting to make `swag-img` as a requirement for `pug` since `swag-img` creates a `builtSwagList` array which updates the image source. My implementation moved the mapping to the top of the file (it's computationally insignificant so there's minimal performance loss when running non-build tasks) so all of the tasks can run in parallel

That being said, based on my implementation, `swag-img:build-data` could be removed since reading from `data.json` is now only done once at the top of the file but I'd rather do that separately

swapagarwal:
This is lit! 🔥

plibither8:
Wait! There is an issue, sorry for not pointing it out earlier, just saw it
:|

Navigating to a URL with the `tags` parameter set **does not** filter out
the content, though it sets the tag values in the Selectr checklist.

vikaspotluri123:
For now the fix is to run `renderSwag` onload

plibither8:
That's the only fix I see.

...unless we build the site on a proper server (like ExpressJS) and host it on heroku thereby allowing complete server-side freedom. Just thinking out loud :smile:

vikaspotluri123:
I think it's perfectly fine to re-render for filtering, especially once the todos are finished

plibither8:
Just pushed a commit that fixes this :+1:

PR #87, opened 2018-10-05
Change link to devswag.io to https

mc388:

(Something else)  

Changes link in readme to https because it is nicer to offer a https domain than a http domain and it's already redirecting from http to https


swapagarwal:
Hey Marvin, thanks for taking the time to create a Pull Request! 😇
However, this decision to handle redirection at server-side was taken to have more control in case we have to move out of Netlify one day.

Issue #92, opened 2018-10-07
Performance Improvements


vikaspotluri123:
I'm in the progress of implementing performance improvements both to the backend and frontend. Here's what's being done:

- [x] minify html at buildtime
- [x] use gulp@next vs github:gulp#4
- [x] load scripts at document end vs head
- [x] remove `swag-img:build-data` task (processing is done on gulpfile load)
- [ ] Normalize formatting + function declarations
- [ ] Add cachebusting + extended caching
----
- [ ] filter: toggle visibility based on classes to reduce re-rendering requirement
- [ ] sort: Clone DOM nodes rather than use template literals. This is to DRY up and possibly get performance improvements
- [ ]  Disallow sorting by difficulty when only one difficulty is shown
- [ ] Define minimum image height (currently thinking 275px) to prevent content jumping

Feel free to add any other suggestions, or discuss anything above

plibither8:
These points are great! I believe, and this is just my opinion, that we use a proper web framework like Express. It might seem overkill, but IMHO it will do most justice to the optimisations @vikaspotluri123 mentioned. Moreover, using Gulp to help pre-render data, optimise it, and build the files seems alright, but we can get the same things with much less code and more readability. Thoughts?

vikaspotluri123:
Since the optimizations I'm making are to the gulp build process, express would technically negate them 😛 That's not to say express is a bad idea though 😄 	 

The improvements I'm making will hopefully make the code more readable, albeit slightly bigger in size (let's just say it's stockpiling for winter 😉) and with the current spec I have in mind, will remove the need for the swag in a JSON format.

I personally prefer static over dynamic (ironically I prefer backend over frontend 🤫) because it cuts out a vector for vulnerability, whether it be from contributor-introduced  bugs or an attack vector, and it reduces maintenance costs overall

plibither8:
You do make some good points, though I'd still be implicitly pushing towards a web framework (don't mind me :smile:).

>and with the current spec I have in mind, will remove the need for the swag in a JSON format.
This sounds interesting, could you elaborate?

vikaspotluri123:
> This sounds interesting, could you elaborate?

[...]

PR #93, opened 2018-10-07
⚡Backend Performance improvements

### LINKED ISSUE #146: Netlify builds are failing
Seen in #140 #145

@plibither8 @vikaspotluri123 @swapagarwal 

I think cherry-picking this commit would fix NPM failing to install packages. https://github.com/swapagarwal/swag-for-dev/pull/93/commits/26e36330f63fc1f1eb4b74432b43d5b3333ed49c

Gona have a try.

vikaspotluri123:
- Refs #92 (hopefully will close by the time I'm done)
- Todos are in the issue (I'm managing the issue so it's not too big of a deal)
- This PR is for backend improvements, I want to do frontend improvements separately to keep changes minimal

Closes #146
Closes #147

### DIFF
--- a/site/package.json
+++ b/site/package.json
@@ -20,7 +20,6 @@
     "eslint": "^5.6.1",
     "gulp": "^4.0.0",
     "gulp-babel": "^8.0.0",
-    "gulp-cli": "^2.0.1",
     "gulp-concat": "^2.6.1",
     "gulp-download-stream": "^0.0.19",
     "gulp-htmlmin": "^5.0.1",


aslafy-z:
The gulp pipeline is way cleaner that way ! Thank you @vikaspotluri123 :tada:

vikaspotluri123:
todo: get rid of preload headers

site/get-data.js
```diff
@@ -0,0 +1,30 @@
+const swagImages = [];
+const fileNames = [];
+const swagList = require('../data.json').map(swag => {
+    // Generate unique filename
+    // @todo: do we want to force jpeg as the image format?
```
aslafy-z: I forced jpeg in the initial code to standardize all images, maybe we could in the future create variants like webp to optimize their sizes even more but I'm not sure about keeping the original format if it's what you're asking about

site/get-data.js
```diff
@@ -0,0 +1,30 @@
...
+    // const extension = swag.image.split('.').pop();
```
aslafy-z: Also, this kind of cut is not really safe, it should rely on mimetype instead

site/get-data.js
vikaspotluri123: I wasn't, but I was also thinking about that 😄 

I was asking more along the lines of if the image comes from the source as a PNG, is there a good reason to convert it to a jpeg (with the implication file type was actually considered by the source)?

site/get-data.js
vikaspotluri123: True, but from what I can tell that adds quite a bit of complexity to the build system. While there could be a CI check (i.e. make sure the mime types match the extension) the complexity cost isn't worth it in the gulpfile

aslafy-z: I agree, that's why jpeg was forced

vikaspotluri123: We could also manually specify the format in the data file and default to jpeg if it's not provided 🤔

site/gulpfile.js
```diff
@@ -1,137 +1,163 @@
...
-                }]
-            ]
-        }))
+        .pipe(babel({ presets }))
         .pipe(uglify())
         .pipe(gulp.dest('dist/assets/js'));
 });
 
+gulp.task('swag-img:clean', () => {
+    return del('dist/assets/swag-img/**/*');
```
aslafy-z: This path could lose two :star:s

site/gulpfile.js
```diff
@@ -1,137 +1,163 @@
...
-            },
-        }, {
-            quality: 90,
-            progressive: true,
-            compressionLevel: 9,
-            errorOnEnlargement: false,
-            errorOnUnusedConfig: false,
-        }))
+        .pipe(responsive([{
+            name: '**/*',
```
aslafy-z: Two more :star:s to bust

site/gulpfile.js
```diff
@@ -1,137 +1,163 @@
...
+gulp.task('webserver', () => {
+    return gulp.src('dist')
+        .pipe(webserver({
+            livereload: true,
+            open: true
+        }));
+});
 
 gulp.task('watch', () => {
     gulp.watch('src/pug/**/*.pug', gulp.parallel('pug'));
```
aslafy-z: Here again, :star:s

Sorry for the flood :fearful:

vikaspotluri123: This would need to be recurisve, but not js; if you change includes/body, pug needs to rebuild

vikaspotluri123:
image
Did someone say caching 😎

site/gulpfile.js
```diff
@@ -1,137 +1,163 @@
...
-    const tags = Array.from(builtSwagList.reduce(
+    const tags = Array.from(swagList.reduce(
         (tagList, { tags }) => {
             tags.forEach(tag => tagList.add(tag));
             return tagList;
         },
         new Set()
     ));
 
+    const manifest = require('./rev-manifest');
```
aslafy-z:
```suggestion
    const manifest = require('./rev-manifest.json');
```

as below.

site/gulpfile.js
```diff
@@ -1,137 +1,163 @@
...
+        'dist/assets/swag-img/*'
+    ]);
 });
 
-gulp.task('swag-img', gulp.parallel('swag-img:build-data', gulp.series('swag-img:clean', 'swag-img:download' , 'swag-img:optimize')));
+gulp.task('cachebust', cb => {
+    const bustedFiles = [
+        'dist/assets/css/*',
+        'dist/assets/js/*',
+        'dist/assets/swag-img/*'
```
aslafy-z: This needs to be done for `dist/assets/img` also, right ?

aslafy-z: ok lgtm till we have more (if we have more)

aslafy-z:
This is ready to deploy right? @vikaspotluri123

vikaspotluri123:
Now there should be no blocker

aslafy-z:
Actually, I think having `gulp-cli` installed locally is ok. It doesn't require users to install `gulp` globally before coding. `npm run build` is made possible because `gulp-cli` is installed.

vikaspotluri123:
`npm run build` should use the module that ships with gulp 🤔 	

Either way, if we want to do that, I think it should be done as a separate PR since that's outside the scope of this one

aslafy-z:
Actually you deleted the gulp-cli dependency but haven't added it. So removing the dep is outside of the scope of this PR.

Gulp does not provide any cli within the gulp package, AFAIK. Correct me if I'm wrong

vikaspotluri123:
You're right 😳 

Looking at Gulp's deps: https://github.com/gulpjs/gulp/blob/v4.0.0/package.json#L34

No need to worry 😄 Since we're not using the CLI explicitly, we don't need it as a dependency 👍

aslafy-z:
I was wrong! I thought we had to install it anyway. Good thing we don't have to! :+1: 

So, no more blocker right?

vikaspotluri123:
From my side, we're good 👍

aslafy-z:
Nice! Could you rebase and merge?

vikaspotluri123:
I can rebase and merge, but haven't we been doing squashes?

plibither8:
Yup, squash and merge.

vikaspotluri123:
:shipit:

PR #105, opened 2018-10-09
Update index.html

vishal6404:

Fill the corresponding section below:

(New Swag Proposal)

* Company:
* Reference URL: 

Q. Have you received swag from this?  
Q. How long did it take?  
Q. Does it ship worldwide?  

(Something else)  

### DIFF
--- a/index.html
+++ b/index.html
@@ -4,7 +4,7 @@
    <meta http-equiv="refresh" content="0; url=http://devswag.io">
 </head>
 <body>
-   <p>Redirecting to: 
-   <a href="http://devswag.io">devSwag</a></p>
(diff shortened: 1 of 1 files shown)


aslafy-z:
Could you please fill the PR template and set a proper title ?
What are you trying to achieve ?

Why `<marquee>` ? It's an  [obselete](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/marquee) tag that makes the content scroll.
There is no need to scroll tis text, this redirection page should be the tiniest as possible.

aslafy-z:
Please reopen with an explanation if you think it needs more attention. Thank you.

Issue #94, opened 2018-10-07
Show tags alphabetically in dropdown


swapagarwal:
Currently, `accessories` is shown towards the end while it should be at the start of the list.

vikaspotluri123:
Hint: look at `site/src/pug/includes/body.pug` to see how the tag list is loaded, and make necessary changes to sort the tag list alphabetically

swapagarwal:
@alekangelov Go ahead!

swapagarwal:
@alekangelov Are you working on this?

PR #107, opened 2018-10-10
fix issue #94

greybax:

Is it regarding a new swag proposal or something else?  no
  
This is fix for issue #94

### DIFF
--- a/site/gulpfile.js
+++ b/site/gulpfile.js
@@ -33,9 +33,7 @@ gulp.task('pug', () => {
             return tagList;
         },
         new Set()
-    ));
-    
-    tags.sort();
+    )).sort();
 
     return gulp.src('src/pug/*.pug')
         .pipe(pug({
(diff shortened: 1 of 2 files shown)


swapagarwal:
@greybax Are you sure it's the right issue?

greybax:
@swapagarwal oohh, sorry, seems that I've wrong marked the issue which I've fixed. Of course it should be issue #94

greybax:
@vikaspotluri123 yeah, seems its better solution. Will fix it now.

vikaspotluri123:
You can just chain to the initial logic

```js
const tags = Array.from(builtSwagList.reduce(
        (tagList, { tags }) => {
            tags.forEach(tag => tagList.add(tag));
            return tagList;
        },
        new Set()
    )).sort();
```

swapagarwal:
Thanks @greybax and @vikaspotluri123! 👍

PR #109, opened 2018-10-10
Sort tags after they get loaded

### LINKED ISSUE #94: Show tags alphabetically in dropdown

pradyunsg:
Fixes #94


vikaspotluri123:
duplicate of #107

PR #111, opened 2018-10-10
Add LICENCE file

aslafy-z:
According to https://github.com/swapagarwal/swag-for-dev#license section

Taken from here: https://github.com/licenses/license-templates/blob/master/templates/cc0.txt


swapagarwal:
Thanks @aslafy-z! 👍

PR #118, opened 2018-10-12
Added gulp build to README

RajatArora08:
- [X] I've checked that this isn't a duplicate pull request.

Is it regarding a new swag proposal or something else?  Something else.
Fill the corresponding section below:

(New Swag Proposal)

* Company:
* Reference URL: 

Q. Have you received swag from this?  
Q. How long did it take?  
Q. Does it ship worldwide?  

(Something else)  
Please describe your changes here: 
Adds gulp build to README #117

PR #121, opened 2018-10-12
Fixes gulp default task. #118

RajatArora08:
Steps to reproduce the issue (as per README):
1. Fresh clone.
2. `cd site`
3. `npm install`
4. `gulp`

The above leads to an error `Error: File not found with singular glob: ~/swag-for-dev/site/dist`.

This is most probably because gulp starts the webserver before the files are built since the tasks are in parallel. Apart from this, even if `gulp build` is run before `gulp`, it displays a blank page on localhost:8000, until manually refreshed.

Changing gulp default in the series `build --> webserver --> watch` fixed the issue.

### DIFF
--- a/site/gulpfile.js
+++ b/site/gulpfile.js
@@ -160,4 +160,4 @@ gulp.task('build', gulp.series(
     )
 ));
 
-gulp.task('default', gulp.parallel('webserver', 'build', 'watch'));
+gulp.task('default', gulp.series('build', 'webserver', 'watch'));


RajatArora08:
@aslafy-z Until your Docker PR is merged, this might fix the issue. Give it a try on a fresh clone.

aslafy-z:
**WORKAROUND**: Creating the `dist` folder before all solves the issue.

Issue #104, opened 2018-10-09
Add Applause-button to swag items


plibither8:
It would be a simple and nice feature wherein a small applause button accompanies a swag item. Users can '+1' it if they feel this is a nice opportunity, and sorting can then also be on the basis of the number of applauses an item has got.

Implementation of this is fairly simple: https://applause-button.com/. This allows for Medium-style applause buttons on static websites with a zero-configuration setup.

swapagarwal:
Neat idea!

swapagarwal:
As the number of claps comes from the server, how would we embed it into the items array to allow sorting?

plibither8:
We'll use the same API method the source code is using: https://github.com/ColinEberhardt/applause-button/blob/master/src/applause-button.js

It'll require a few AJAX calls to fetch the number of claps, and after that it's simple sorting 😄

plibither8:
Better yet: just one ajax request will do: https://github.com/ColinEberhardt/applause-button-server/blob/master/src/getMultiple.js

swapagarwal:
Looks good! Do you want to pick this up?

plibither8:
I won't get much time to work on this now because my exams are approaching
:/. I'd love to review or help out from time to time though 😀

swapagarwal:
No issues. :)

caseydierking:
@swapagarwal  I'll pick this up.

caseydierking:
Relinquishing this. Won't be able to get to it.

giordifungula:
I can take this on :)

pannapat:
Is this still available to take on?

aslafy-z:
Hello @pannapat, it's yours! I'll assign you. We have a stale PR at https://github.com/swapagarwal/swag-for-dev/pull/124, please take a look.

pannapat:
Thanks a lot! Please allow me to look into this for a couple of days.

itguymax:
@anhek  the swag-for-dev website is not updated i thought to contribute to some features. anyone here?

itguymax:
is this still open?

Yash-Garg:
Great idea 👍

pannapat:
I’ve been working on this issue! Sorry for not keeping you all posted

pannapat:
> Hello @pannapat, it's yours! I'll assign you. We have a stale PR at #124, please take a look. You might need to synchronize with @anhek since the redesign (#443) will happen soon!

[...]

PR #124, opened 2018-10-16
Add applause button

andrewda:

This adds an applause button to allow users to vote on their favorite swag!

Closes https://github.com/swapagarwal/swag-for-dev/issues/104

image

## Still TODO

- [x] Allow sorting

### DIFF
--- a/site/package-lock.json
+++ b/site/package-lock.json
@@ -3491,15 +3491,13 @@
           "version": "1.0.0",
           "resolved": false,
           "integrity": "sha1-ibTRmasr7kneFk6gK4nORi1xt2c=",
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "resolved": false,
           "integrity": "sha512-iCuPHDFgrHX7H2vEI/5xpz07zSHB00TpugqhmYtVmMO6518mCuRMoOYFldEBl0g187ufozdaHgWKcYFb61qGiA==",
           "dev": true,
-          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
@@ -3516,22 +3514,19 @@
           "version": "1.1.0",
           "resolved": false,
           "integrity": "sha1-DQcLTQQ6W+ozovGkDi7bPZpMz3c=",
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "concat-map": {
           "version": "0.0.1",
           "resolved": false,
           "integrity": "sha1-2Klr13/Wjfd5OnMDajug1UBdR3s=",
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "console-control-strings": {
           "version": "1.1.0",
           "resolved": false,
           "integrity": "sha1-PXz0Rk22RG6mRL9LOVB/mFEAjo4=",
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "core-util-is": {
           "version": "1.0.2",
@@ -3662,8 +3657,7 @@
           "version": "2.0.3",
           "resolved": false,
           "integrity": "sha1-Yzwsg+PaQqUC9SRmAiSA9CCCYd4=",
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "ini": {
           "version": "1.3.5",
@@ -3677,7 +3671,6 @@
           "resolved": false,
           "integrity": "sha1-754xOG8DGn8NZDr4L95QxFfvAMs=",
           "dev": true,
-          "optional": true,
           "requires": {
             "number-is-nan": "^1.0.0"
           }
@@ -3694,7 +3687,6 @@
           "resolved": false,
           "integrity": "sha512-yJHVQEhyqPLUTgt9B83PXu6W3rx4MvvHvSUvToogpwoGDOUQ+yDrR0HRot+yOCdCO7u4hX3pWft6kWBBcqh0UA==",
           "dev": true,
-          "optional": true,
           "requires": {
             "brace-expansion": "^1.1.7"
           }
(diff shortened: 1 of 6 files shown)


zac-garby:
In my opinion the applause button is too big. It looks a bit out of place. I think maybe it should be the same size as the difficulty indicator circle (which could be enlarged slightly too)

andrewda:
@zac-garby Agreed - only difficulty is that the numbers start to overlap with the button at anything below 42px. I'll try bumping up the size of the difficulty indicator and we'll see if that makes it any better.

plibither8:
This looks great!

There were two things that came to mind:
1. The API URLs should not use the name of the company/organisation of the swag item. There can be instances where the same company has had multiple swag opportunities (for example, Digital Ocean). An ideal fix to this would be to add a timestamp along with the swag item name, and the timestamp would be of the date the opportunity was created or added to the list.
2. Before pushing these changes, is there a way to set all claps back to 0?

swapagarwal:
A weird bug: Clap 10 items for an item and reload the page. The last clap isn't persisted. 🤔

On an unrelated note, should we restrict the user to one clap per item?

plibither8:
:+1: for limiting to one clap

site/package-lock.json
```diff
@@ -3491,13 +3491,15 @@
           "version": "1.0.0",
```
aslafy-z: You should checkout this file. Here's the related npm issue: https://github.com/npm/npm/issues/17722

site/src/js/index.js
```diff
@@ -4,13 +4,39 @@
...
+const API = 'https://api.applause-button.com';
+
+const requestClaps = (urls) => {
+    return new Promise((resolve) => {
+        fetch(`${API}/get-multiple`, {
+            method: 'POST',
+            headers: {
+                'Content-Type': 'application/json; charset=utf-8',
+            },
+            referrer: 'devswag.io',
```
aslafy-z: The `Referer` HTTP header needs an URL (with it's http* prefix): https://tools.ietf.org/html/rfc1945#section-10.13

Not sure it's really needed as it looks to be only used as a key.

site/src/js/index.js
```diff
@@ -29,6 +55,8 @@ const renderSwag = () => {
             case 'Alphabetical, reversed': return a.name.toLowerCase() < b.name.toLowerCase() ? 1 : -1;
             case 'Difficulty':             return difficultyIndex(a.difficulty) > difficultyIndex(b.difficulty) ? 1 : -1;
             case 'Difficulty, reversed':   return difficultyIndex(a.difficulty) < difficultyIndex(b.difficulty) ? 1 : -1;
+            case 'Popular':                return claps[`devswag.io/${a.name}`] < claps[`devswag.io/${b.name}`] ? 1 : -1;
```
aslafy-z: You may want to move the `devswag.io` prefix to a constant string, to reduce repetition.

site/src/pug/includes/body.pug
```diff
@@ -25,7 +25,7 @@ include ./fork.pug
                     .select-wrapper
                         label.sr-only(for="sorting") Sort opportunities
                         select#sorting
-                            for filter in ['Alphabetical', 'Alphabetical, reversed', 'Difficulty', 'Difficulty, reversed']
+                            for filter in ['Alphabetical', 'Alphabetical, reversed', 'Difficulty', 'Difficulty, reversed', 'Popular', 'Popular, reversed']
```
vikaspotluri123: This should be programmatically applied after the claps come in from the API - otherwise the logic for sort-by-popularity won't work

i.e.
```js
requestClaps(urls).then(resp => {
	claps = response;
	const rNode = document.createElement('option');
	node.value = 'popularity';
	node.textContent = 'Popularity';
	// ... create the node for reverse
	selectInput.appendChild(node);
	selectInput.appendChild(rNode);
});

File: site/src/pug/includes/head.pug [hunk 5]
```diff
@@ -14,6 +14,9 @@ link(rel="apple-touch-icon" size="128x128" href="/assets/img/logo.png")
 link(rel="stylesheet" href="/assets/css/index.css")
 link(rel="stylesheet" href="https://fonts.googleapis.com/css?family=Cabin")
 
+link(rel="stylesheet" href="https://unpkg.com/applause-button/dist/applause-button.css")
```
vikaspotluri123: imo Selectr is more important than applause so it should be loaded before applause

aslafy-z:
@plibither8 @andrewda A solution (for the first point you rose) would be to use the actual reference url instead of `devswag.io/:swagname`. To avoid conflicts with others claps buttons we may want to prefix the url by a "devswag proxy": `devswag.io/claps/:swagurl`.

plibither8:
Good idea @aslafy-z!

site/src/js/index.js
```diff
@@ -85,6 +117,25 @@ const UrlHandler = () => {
 window.addEventListener('load', () => {
     UrlHandler();
 
+    const urls = window.swag.map(getItemUrl);
+
+    requestClaps(urls).then(resp => {
+        claps = resp;
+
+        const selectInput = document.getElementById('sorting');
```
vikaspotluri123: You can delete this line since `sortingInput` is a global with the same value

site/src/pug/includes/head.pug
aslafy-z: I think this link & script have to be moved to the end of the body (link staying before the script tag).

> `<link rel="stylesheet">` is **allowed** in the body: https://html.spec.whatwg.org/multipage/links.html#body-ok

site/package-lock.json
aslafy-z: This file still needs to be checkout.

swapagarwal:
@andrewda Can you rebase this on master?

swapagarwal:
Pinging @andrewda

Issue #88, opened 2018-10-05
Suggestion: Don't remove items from data.json but set `active: false`


plibither8:
If an opportunity for free swag has expired, I suggest we don't remove the item from the `data.json` file. This will allow a user to browse through past opportunities and offers, some of which might even restart.

In light of that I propose either of the two implementations:

1. If an item has expired, add an `expired: true` property to the item. Otherwise, set it to false.

2. Divide the list into two portions: one with active opportunities and the other with expired/inactive ones.

aslafy-z:
I vote for the `expired: true` way of doing it. We would be able to filter swag opportunities from the gulp pipeline if we dont want them on the website. 
Although i'm not sure `expired` is the best label, maybe `active` whould be better?

plibither8:
Agreed, `active` is better as `inactive` is more befitting.

swapagarwal:
+1 for `active: true/false`

swapagarwal:
Next Steps:
1. Add `active` key in `data.json`
1. Filter out inactive items in gulp pipeline

Seems straight-forward. Is there any newcomer that wants to work on this? We'll help you along the way! 😇

sank20:
Hey, I'm new here, I would like to take up this issue, if I may. To confirm, I need to add the `active` key in each json object. Do I need to check whether the item is active and provide the correct value or just some default value?
I'm not familiar with gulp. Can you please guide me for filtering out inactive items from gulp pipeline?

sank20:
Thanks @plibither8! Cool, so I guess creating a new filter should ideally be a new issue that would be referenced to this one.

plibither8:
No need to create a separate issue, just create a new PR with the changes

sank20:
Alright

swapagarwal:
@sank20 Let us know if you need any help!

rajeshwarirapelli25:
Hey,I'm working on this issue.

sank20:
Hi @rajeshwarirapelli25 ! Thanks for helping me for this as I'm gonna be super busy the whole month. Please note that you also need to add the filter for active/inactive items. Please refer the comments and feel free to ask your doubts here.

aslafy-z:
from @swapagarwal [comment](https://github.com/swapagarwal/swag-for-dev/pull/102#issuecomment-431578623)
> Can think of adding something like `validThru` key to automatically make these `inactive` in November..

This would be great for all hacktoberfest events.

[...]

PR #134, opened 2018-10-19
fixes issue #88 - added key "active" in data.json

rajeshwarirapelli25:
added key "active" with default value as true since all items in the data.json are active


data.json
```diff
@@ -5,126 +5,142 @@
     "description": "Publish a skill, get a hoodie (varies by country) and an Echo Spot (if 1,000 unique customers use your skill in its first 30 days of publication)!",
     "reference": "https://developer.amazon.com/alexa-skills-kit/alexa-developer-skill-promotion",
     "image": "https://images-na.ssl-images-amazon.com/images/G/01/mobile-apps/dex/alexa/alexa-skills-kit/ASM0761-Dev-LP-Goodiepage-UK-March._CB501585903_.png",
-    "tags": ["clothing", "device"]
+    "tags": ["clothing", "device"],
+    "active":true
```
aslafy-z: There should be a space between the ':' and true.

vikaspotluri123:
Do we want to spec out how the active tag should be used for the site - e.g. filter active items in gulp?

aslafy-z:
Follow the discussion here: #88

PR #140, opened 2018-10-20
Add collaborators

swapagarwal:
- [x] I've checked that this isn't a duplicate pull request.

(Something else)  
Please describe your changes here: This PR adds collaborators to the footer of devswag.io

@aslafy-z (I was able to contact @vikaspotluri123 and get his thoughts regarding this but I couldn't find your contact details so tagging you here...)
Would you like your name to be included in the footer of devswag.io? (something like maintained by xyz...) 🎉
P.S. I expect the maintenance effort to decrease substantially once Hacktoberfest ends! 💯 
Feel free to modify the link you wish to be redirected to when users click on it. 😇

PR #145, opened 2018-10-23
Fix: Minor style issues

plibither8:

Two minute changes:

1. The left-edge of the first tag did not align properly with the edge of the card due to excess margins. I have added negative margins to the parent element (`.item`).

2. On mobile devices, the feedback button on the right edge is positioned such that it causes horizontal scrolling. I have disallowed scrolling. This might seem like a hack-y way to do it but it works.

### DIFF
--- a/site/src/styl/imports/card.styl
+++ b/site/src/styl/imports/card.styl
@@ -34,7 +34,7 @@
         color #222
 
         &.swag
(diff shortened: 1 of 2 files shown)


vikaspotluri123:
Similar netlify issue in #140 - I won't get a chance to debug until the weekend

PR #153, opened 2018-10-31
Added the "active" key to all that sets! 

lovjs:
However, I did not have time to add a filter for it. ):

- [ ] I've checked that this isn't a duplicate pull request. It is similar but the previous one didn't add the key to all of the sets just one.


aslafy-z:
Follow the discussion here: #88

PR #156, opened 2018-11-05
HOTFIX: Filter out hacktoberfest data

vikaspotluri123:
As we don't have support for client-side filtering based on swag state, this is a hotfix to disable all hacktoberfest-related swag from being shown. Once we add the state filtering, this can be removed


swapagarwal:
looks good to me

vikaspotluri123:
Now that we have more contributors, what do you think of having (as a general rule of thumb) at least 2 review approvals for a PR to get merged @swapagarwal 	?

PR #154, opened 2018-11-01
Removing this entry as Hacktoberfest 2018 is over and the offering has ended

nul800sebastiaan:
- [x] I've checked that this isn't a duplicate pull request.

PR #76, opened 2018-10-03
feat: add dockerfile 🐳

aslafy-z:
Add a Dockerfile in order to ease website development with fixed dependencies and tooling. 

Two variants:
- `devswag:dev`: development image, does not contain any code but all the tooling to work with it
- `devswag`: production image, uses nginx to serve the built artifacts

More infos: https://github.com/aslafy-z/swag-for-dev/blob/add-dockerfile/CONTRIBUTING.md#lets-start-hacking

I just pushed the two variants to my docker-hub if anyone wants to try them without downloading the code (can't use dev images as it needs to be plugged to the code using volumes): https://hub.docker.com/r/zadki3l/devswag/tags/

**TODO**
- [ ] You tell me

PR #159, opened 2018-11-11
Create `expired` tag to mark expired promotions

### LINKED ISSUE #88: Suggestion: Don't remove items from data.json but set `active: false`

### LINKED ISSUE #158: Remove Passenger

The Passenger promotion has ended (since the livestream was on November 1st), and so it should be removed from the website.
I can open a PR if there are no objections.

itaisteinherz:

This PR is similar to #134 and #153, but the concept is a bit different.
Continuing the discussion at #88, it seems to me like a good option for now would be to mark expired promotions with the `expired` tag, and then filter those out.
That is what this PR adds.

We could also use a `expiration` property to set the date and time of expiration of a promo, similarly to what was suggested in https://github.com/swapagarwal/swag-for-dev/issues/88#issuecomment-431614021 and https://github.com/swapagarwal/swag-for-dev/issues/88#issuecomment-431579004.

Fixes #88
Fixes #158
Closes #134
Closes #153
Closes #154

// cc @vikaspotluri123


itaisteinherz:
> we just used the Readme for the site, we had a section for expired swag. I think we should do the same here

That seems like a bad idea to me, since this way even expired promos are displayed on the website, and the user has to go the the website's readme to know which promos are currently active.

vikaspotluri123:
Sorry if I wasn't clear, what I meant is the idea to add a separate section / tab which contains all of the expired opportunities as we had in the past. This would be on the site as well :)

itaisteinherz:
We could allow the user to choose the `expired` tag to view expired promos explicitly, since the PR already adds the tag to the appropriate expired promos.
We could also add mention the expired promos in the readme, but that would probably get out-of-sync with `data.json` very quickly, and would be difficult to maintain.

aslafy-z:
I'd go for keeping expired items to the swag list and adding additional logic to the frontend filtering to show expired items only when the `expired` is explicitly selected. Would this fits you?

vikaspotluri123:
> We could also add mention the expired promos in the readme, but that would probably get out-of-sync with `data.json` very quickly, and would be difficult to maintain.

The readme is no longer in active use, it's a static file used to provide info about the repo and link to devswag.io

> We could allow the user to choose the `expired` tag to view expired promos explicitly, since the PR already adds the tag to the appropriate expired promos.

For me personally, I need to see a design or implementation to make a judgment - there are lots of different ways to implement something, but many of them can have bad UI / UX, which is my main focus 😄

itaisteinherz:
> I'd go for keeping expired items to the swag list and adding additional logic to the frontend filtering to show expired items only when the `expired` is explicitly selected

This seems like the best option to me.
@vikaspotluri123 I could implement it and together we can iterate over the implementation details afterwards.

vikaspotluri123:
Sounds good 👍

itaisteinherz:
0402a5dab606c58fa1c16c17977e468ec1c4e9d3 should have implemented @aslafy-z suggestion, but when running `gulp` locally I couldn't see the tag selector after the first time the command was run (I suspect this is related to the manifest file and [`gulp-rev`](https://github.com/sindresorhus/gulp-rev)), and so I couldn't really test out the changes.
Let me know if it works for you.

aslafy-z:
Here's the preview url for this PR: https://deploy-preview-159--peaceful-chandrasekhar-efde8e.netlify.com/

Looks like the expired stuff appears by default. 

Also, should we grey out the expired swag items blocks using css or something ?

site/package.json
```diff
@@ -4,6 +4,7 @@
   "description": "swag opportunities for developers",
   "main": "index.js",
   "scripts": {
+    "start": "gulp",
```
aslafy-z: Please remove this NPM script, #76 gonna add it :wink:

itaisteinherz:
> Also, should we grey out the expired swag items blocks using css or something?

I tried to hide the expired items by default, but for some reason they continue to appear.

site/package.json
itaisteinherz: Done in e9988539953c94a6a4e90283c48f670014c1d9ef 👌🏻

vikaspotluri123:
116 has been merged!

itaisteinherz:
Note that the build is failing because of one of the promo images being unavailable (I've opened #181, which fixes this).

aslafy-z:
Just in case: To trigger the build again, you can create an empty commit with 
```
git commit -m 'Trigger build' --allow-empty
```

vikaspotluri123:
@itaisteinherz if you're having any issues with merge conflicts, let me know, I'll do my best to help out 🤗

site/src/js/index.js
```diff
@@ -52,7 +52,8 @@ function handleTags() {
 
     Array.from(contentEl.querySelectorAll('.item')).forEach(el => {
         const show = tags.reduce((sho, tag) => sho || el.classList.contains(`tag-${tag}`), false);
-        if (!show) {
+
+        if (!show || (![...tags].includes("tag-expired") && el.classList.contains("tag-expired"))) {
```
aslafy-z: is the array copy really needed to call `includes`?

vikaspotluri123: This logic should be in the reduce logic because that's what's determining if the card should be visible or not 🤔

itaisteinherz: The `expired`-logic shouldn't be in the `reduce` since it only needs to happen once. I'll move it to line 54 though, since the `expired` filtering should affect `show`.

vikaspotluri123: i.e.

```js

const show = tags.reduce(fn, tags.includes('tag-expired') && el.classList.contains('tag-expired'))

```

vikaspotluri123: You're right 🤔 we'll table this for now - if we come up with something better by the time this is ready to be merged, we can come back to it, otherwise, we can merge with this 🙂

aslafy-z:
Made some tests:

- [no tags are selected](https://deploy-preview-159--peaceful-chandrasekhar-efde8e.netlify.com/): Expired swag is shown (should be hidden)
- [`expired` tag is selected](https://deploy-preview-159--peaceful-chandrasekhar-efde8e.netlify.com/?tags=expired): No expired swag shown (should be shown)
- [`hacktoberfest` tag is selected](https://deploy-preview-159--peaceful-chandrasekhar-efde8e.netlify.com/?tags=hacktoberfest): No expired swag shown (should be shown after other items?)
- [`hacktoberfest` and `expired` tags are selected](https://deploy-preview-159--peaceful-chandrasekhar-efde8e.netlify.com/?tags=hacktoberfest+expired): No expired swag shown (show be shown after other items?)

site/src/js/index.js
```diff
@@ -51,7 +51,8 @@ function handleTags() {
     }
 
     Array.from(contentEl.querySelectorAll('.item')).forEach(el => {
-        const show = tags.reduce((sho, tag) => sho || el.classList.contains(`tag-${tag}`), false);
+        const show = tags.reduce((sho, tag) => sho || el.classList.contains(`tag-${tag}`), false) && !(![...tags].includes("tag-expired") && el.classList.contains("tag-expired"));
```
aslafy-z: I don't think copying the array is necessary. 

```suggestion
        const show = tags.reduce((sho, tag) => sho || el.classList.contains(`tag-${tag}`), false) && !(!tags.includes("tag-expired") && el.classList.contains("tag-expired"));
```

itaisteinherz:
> [`hacktoberfest` tag is selected](https://deploy-preview-159--peaceful-chandrasekhar-efde8e.netlify.com/?tags=hacktoberfest): No expired swag shown (should be shown after other items?)

I think that in this case no expired swag should be shown.

> [`hacktoberfest` and `expired` tags are selected](https://deploy-preview-159--peaceful-chandrasekhar-efde8e.netlify.com/?tags=hacktoberfest+expired): No expired swag shown (show be shown after other items?)

Here, only expired swag tagged `hacktoberfest` should be shown.

site/src/js/index.js
```diff
@@ -51,7 +51,8 @@ function handleTags() {
     }
 
     Array.from(contentEl.querySelectorAll('.item')).forEach(el => {
-        const show = tags.reduce((sho, tag) => sho || el.classList.contains(`tag-${tag}`), false);
+        const show = tags.reduce((sho, tag) => sho || el.classList.contains(`tag-${tag}`), false) && !(!tags.includes("tag-expired") && el.classList.contains("tag-expired"));
```
vikaspotluri123:
```suggestion
        const show = (tags.includes("tag-expired") && !el.classList.contains("tag-expired")) &&
        	tags.reduce((sho, tag) => sho || el.classList.contains(`tag-${tag}`), false);
```

itaisteinherz: I committed this in c2f7ab4c7ee4875a13426b9067293e79734385de with a small modification: I used `||` instead of `&&` in the `expired` logic, as dictated by [De Morgan's laws](https://en.wikipedia.org/wiki/De_Morgan%27s_laws).

itaisteinherz:
Actually, I like that idea. It would simplify things both for the user and for us (in development). I'll try to implement that today.

swapagarwal:
@itaisteinherz Needs rebase.

itaisteinherz:
I prefer to avoid using force-pushes to resolve conflicts, and things got a bit messy 😅 but I got it sorted out.

aslafy-z:
I think you lost some changes doing so :/
Any chance you finish this promptly? This needs to be sorted out!

vikaspotluri123:
I found the perfect way to **not** do it

image

<small>okay, I'm done being sarcastic in this repo 😝</small>

plibither8:
Ahahahaha love it! /s

plibither8:
The PR looks good to me, let's merge? @vikaspotluri123 @aslafy-z @swapagarwal

src/js/index.js
```diff
@@ -54,7 +55,8 @@ function handleTags() {
 	}
 
 	Array.from(contentEl.querySelectorAll('.item')).forEach(el => {
-		const show = tags.reduce((sho, tag) => sho || el.classList.contains(`tag-${tag}`), false);
+		const show = (showExpired || !el.classList.contains('tag-expired')) &&
```
aslafy-z: you have to use `showExpired.checked`

aslafy-z:
@plibither8  Right now, the `Show expired opportunities` does not work. It doesn't even retrieve swag using the `expired` tag on my end.

src/pug/includes/body.pug
```diff
@@ -31,9 +31,12 @@ include ./fork.pug
                             option(value="DIFFICULTY_DESCENDING" class="difficulty") Difficulty, reversed
                     span order
 
-            .bottom
+            .bottom.flex
```
vikaspotluri123: We need to merge this into the previous row. It was acceptable (although not really okay) when there was only one option, but that's not the case anymore.

This PR should ref #92 as that's one of the things to do

Things to check

- [ ] wraps properly on smaller screens

vikaspotluri123:
You're looking for the [flex-wrap](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-wrap) css property 😁

src/js/index.js
itaisteinherz: I thought I've already addressed this, but for some reason this hasn't been fixed yet. Thanks for reminding me about this.

itaisteinherz: Fixed in 9a9e23925c9a109afa6e292b7c556cb5a13c44e2.

vikaspotluri123:
@itaisteinherz 	let me know when you would like another review! There's still some stuff that's not working, and I don't want to tell you if you already know! 😊

src/pug/includes/body.pug
itaisteinherz: I've tried to fix this in 5e26d87ef0c5c7a3a2d1328183b98bd0f5f1d66b, but for some reason when building locally the wrapping didn't seem to work properly. I'm probably doing the Stylus selector and selector nesting wrong.
It would be great if you could take a look and let me know what you think should to be changed.

itaisteinherz:
@vikaspotluri123 Just a friendly reminder for you to check out [my comment](https://github.com/swapagarwal/swag-for-dev/pull/159#discussion_r241562604) when you have the time to 🙃 (it's currently one of the main things blocking this PR).

src/pug/includes/body.pug
vikaspotluri123: Thanks for the reminder, this completely fell off my radar! The layout is 100% broken now, so I'm going to go through your commits and see which one broke it, then we can work from there 😇

vikaspotluri123:
> It's weird the build passes. We should add some check as the css generation fails..

Building didn't fail, linting did! This is older than #224 so it wasn't caught 🎣

After #228 merges, CI will fail as well

@itaisteinherz I pushed a commit that fixed the linting issues, was there anything else?

itaisteinherz:
Thanks for the fix!
I think we still need to test the current filtering functionality some more, but other than that this PR is ready to be merged.

itaisteinherz:
I've fixed the expired-related issues in 644ce35db99611884a4e274e6bfc745915a4669f. The code isn't perfect, but it should good for now. I'll fix the conflicts and then this is ready to be merged 👍🏻

itaisteinherz:
[The preview](https://deploy-preview-159--peaceful-chandrasekhar-efde8e.netlify.com/) seems to work quite well, but I may have broken a few things in `body.pug` and `index.styl` while resolving the conflicts. Other than that, this is good to go.

src/js/index.js
```diff
@@ -50,17 +51,17 @@ function handleTags() {
...
-		const show = tags.reduce((sho, tag) => sho || el.classList.contains(`tag-${tag}`), false);
+		const show = (showExpired.checked || !el.classList.contains('tag-expired')) &&
+			(tags.reduce((sho, tag) => sho || (el.classList.contains(`tag-${tag}`) && tag !== 'expired'), false) || tags.length === 0);
 		if (!show) {
 			el.classList.remove('visible');
 		}
 	});
 
-	if (!search) {
+	if (!search || !search.entries().next().value) {
```
vikaspotluri123: This needs to be more verbose as to what it's doing - I think the get method might be more useful here

itaisteinherz:
> Arrows glitch on mobile, it may need some positioning fix

@aslafy-z I tried fixing this but I'm not sure how to. It would be great if any of you could help me with this.

src/styl/index.styl
```diff
@@ -70,9 +70,8 @@ section.main
 			padding 5px 20px 5px 5px
 			-webkit-appearance none
 			-moz-appearance none
-			font-family $default-font-family
```
aslafy-z: Please restore this variable

src/styl/index.styl
```diff
@@ -43,9 +46,6 @@ section.main
 		.item
 			padding 5px
 
-		.select-wrapper
```
aslafy-z: Restoring this fix the select's arrows

src/pug/includes/body.pug
```diff
@@ -71,7 +73,7 @@ include ./fork.pug
 			p Maintained by #[a(href="https://github.com/aslafy-z") Zadkiel] and #[a(href="https://github.com/vikaspotluri123") Vikas Potluri]
 		.item.flex
 			p Want your company's swag to be listed? #[a(href="https://t.me/swapagarwal") Contact us].
-	#codefund_ad
+		#codefund_ad
```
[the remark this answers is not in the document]
itaisteinherz: Nope 😅 I'll restore it.

aslafy-z:
Pretty good! There are two last thing that I'd like to discuss:
- The separator looks too close to the options form
- it looks not so good centered on mobile, we could enlarge each item to fit the page width, wdyt?

vikaspotluri123:
> * it looks not so good centered on mobile, we could enlarge each item to fit the page width, wdyt?

I don't think the opportunities are 100% width; we should match that width, but yes, `.item` should be "full" width on mobile 👍

gulpfile.js
```diff
@@ -34,7 +34,11 @@ let manifest = {
 gulp.task('pug', () => {
 	const tags = Array.from(swagList.reduce(
 		(tagList, {tags}) => {
-			tags.forEach(tag => tagList.add(tag));
+			tags.forEach(tag => {
```
vikaspotluri123:
```suggestion
 			tags.filter(tag => tag !== 'expired').forEach(tag => tagList.add(tag))
```

I'm not entirely sure how to make suggested changes for multiple lines

itaisteinherz:
> `.item` should be "full" width on mobile

@vikaspotluri123 I'm not sure what you meant here, but I think this is how it currently is. I also added some spacing before the separator.

vikaspotluri123:
@itaisteinherz what I was saying is a bit convoluted, a better idea of what to do is to left-align the button flexbox rather than center-align it

itaisteinherz:
> left-align the button flexbox rather than center-align it

That should be done on mobile only, where you can fit one input in each line, right?

vikaspotluri123:
Yes, that's correct 😄

src/styl/index.styl
```diff
@@ -10,6 +10,12 @@ section.main
 	flex-direction column
 	margin 0 auto
 
+	.controls
+		flex-wrap wrap
+
+		@media screen and (max-width: 440px)
```
vikaspotluri123:
```suggestion
 		@media screen and (max-width: 450px)
```

src/styl/index.styl
```diff
@@ -10,6 +10,12 @@ section.main
...
+			justify-content right
```
vikaspotluri123:
```suggestion
 			justify-content left
```

What was the reason for specifying right here?

itaisteinherz: My bad, I though you said right-align 😅
Though having the controls aligned to the right does feel a bit more natural (to me at least).
Do you still want me to change this?

itaisteinherz: My bad 😅

src/pug/includes/body.pug
```diff
@@ -31,34 +30,37 @@ include ./fork.pug
...
-							source(type="image/webp" srcset=swag.images.webp)
-							source(type="image/jpeg" srcset=swag.images.jpeg)
-							img(src=swag.images.jpeg alt=swag.name + " swag you can get")
-					p.description!=swag.description
-					a(href=swag.reference) Check it out
+				.item.flex
+					input(type="checkbox" id="expired")
+					label(for="expired") Show expired opportunities
+
+			.content.flex#content
```
vikaspotluri123: It looks like the indentation was changed here - was that intentional?

itaisteinherz: Nope

itaisteinherz: I think I fixed this in 5b866aa342ff455c36bd3424aa5bb441f686fa49.

src/js/index.js
```diff
@@ -50,17 +51,17 @@ function handleTags() {
...
-		const show = tags.reduce((sho, tag) => sho || el.classList.contains(`tag-${tag}`), false);
+		const show = (showExpired.checked || !el.classList.contains('tag-expired')) &&
+			(tags.reduce((sho, tag) => sho || (el.classList.contains(`tag-${tag}`) && tag !== 'expired'), false) || tags.length === 0);
 		if (!show) {
 			el.classList.remove('visible');
 		}
 	});
 
-	if (!search) {
+	if (!search || !search.get('tags')) {
```
vikaspotluri123: I don't think this should be here - we always want to update the URL when the taglist changes

src/js/index.js
```diff
@@ -50,11 +51,11 @@ function handleTags() {
 
 	if (tags.length === 0) {
 		history.pushState(null, '', window.location.pathname);
-		return;
 	}
 
 	Array.from(contentEl.querySelectorAll('.item')).forEach(el => {
-		const show = tags.reduce((sho, tag) => sho || el.classList.contains(`tag-${tag}`), false);
+		const show = (showExpired.checked || !el.classList.contains('tag-expired')) &&
```
vikaspotluri123:
```suggestion
		const show = (showExpired.checked || !el.classList.contains('tag-expired')) &&
	(tags.reduce((sho, tag) => sho || el.classList.contains(`tag-${tag}`), false));
```

Again with the multiline suggestion, but 

 - No need to check if the tag is `expired`, because we filter it out in the gulpfile
 - No need to check if the taglist is empty, because if it is, we would have returned previously

itaisteinherz: Sorry for the significant delay, I forgot to commit & push this 😅

itaisteinherz:
@vikaspotluri123 @aslafy-z Anything else you'd like me to take a look at / revise?

src/js/index.js
```diff
@@ -50,11 +51,11 @@ function handleTags() {
 
 	if (tags.length === 0) {
 		history.pushState(null, '', window.location.pathname);
-		return;
```
vikaspotluri123: This return statement shouldn't have been dropped

itaisteinherz: I've removed it since we do want the function to continue executing, since if we `return` here expired promos won't get filtered out in case the checkbox isn't selected.

vikaspotluri123: If that's the case, the `reduce` default value needs to be `tags.length===0` (as it was before the latest commit) - otherwise nothing will be displayed by default

itaisteinherz: Done in 1ae91e384cf37d4a3f2c974e17b889b8c851d3bc.

aslafy-z:
Here's a diff snippet to inline controls on mobile

```diff
diff --git a/src/styl/index.styl b/src/styl/index.styl
index ca71131..a9a9695 100644
--- a/src/styl/index.styl
+++ b/src/styl/index.styl
@@ -46,11 +46,16 @@ section.main
                        width 100%
                        justify-content flex-start
 
+               @media screen and (max-width: 450px)
+                       .item, .select-wrapper, select
+                               width 100%
+
                .item
                        padding 5px
 
                .select-wrapper
                        position relative
+                       margin 0 5px
 
                .select-wrapper:before
                        content '▼'
@@ -68,7 +73,6 @@ section.main
                        margin-right 6px
 
                select
-                       margin 0 5px
                        background #fff
                        border none
                        border-bottom 2px solid #bbb
```

Here's a screenshot of how it renders:

vikaspotluri123:
Maybe increase the padding between the elements when there's wrapping?

aslafy-z:
> Maybe increase the padding between the elements when there's wrapping?

This change (mine) introduces some padding incoherences https://github.com/swapagarwal/swag-for-dev/pull/159/commits/e54711cc2f607f59b6f48b175896c9a763fbfef0#diff-19f71ca6255be73cef2f1612a35e79fdL71

aslafy-z:
The "Show expired opportunities" should trigger some changes in the URL, but it doesn't as the expired tag does not exists anymore. Should we restore it and then filter in the frontend?

itaisteinherz:
The expired checkbox is now reflected in the URL by the `showExpired` search param. I've also fixed the expired promos issue, but the build fails because of #240.

aslafy-z:
I think we should ignore the value of showExpired and if the key exists take it as it's enabled.

I don't like the showEnabled property, we could change it for something shorter like 'expired', wdyt?

itaisteinherz:
> I think we should ignore the value of showExpired and if the key exists take it as it's enabled

We could do that, but that would make the URL look ambiguous and unclear. Maybe we could opt for having `?expired=y` to show expired promos and `?expired=n` to hide those?

itaisteinherz:
@aslafy-z @vikaspotluri123 Ready to merge now? 🙃

aslafy-z:
OpenBazaar should be marked as expired.

I think there is one last point that should be fixed (or maybe later by #197), `expired` shouldn't show in url if it's not checked. Now, when loading the page it adds empty tags and expired=n to url by itself.

aslafy-z:
Better now, still, if I check "show expired opportunities" it sets expired to y and tags to an empty value, ie:
https://deploy-preview-159--peaceful-chandrasekhar-efde8e.netlify.com/?tags=&expired=y

aslafy-z:
I think it infinite loop at some point, I'm not able to load the page in my browser :fearful:

itaisteinherz:
@aslafy-z From looking at [the latest build](https://deploy-preview-159--peaceful-chandrasekhar-efde8e.netlify.com/), it seems to me like the issue is caused by upscri.be and isn't our fault. Locally, this works with and without the upscri.be `iframe` enabled.

vikaspotluri123:
> Locally, this works with and without the upscri.be `iframe` enabled.

Doesn't seem to be an issue with ubscribe because devswag.io loads perfectly fine

aslafy-z:
Just found out it was related to the Grammarly Chrome addon. Once I disabled it, it loads so fast! It might be one of the cause lot of websites where loading slowly on my PC these times.

src/js/index.js
```diff
@@ -46,16 +47,22 @@ function handleDifficulty(difficultyChanged) {
 	return false;
 }
 
-function handleTags() {
-	const tags = selectr.getValue();
+function updateUrl() {
+	const newRelativePathQuery = `${window.location.pathname}?${search.toString()}`;
```
aslafy-z: This function can be "optimized" this way:

```js
	let nextPath = window.location.pathname;

	const queryString = search.toString();
	if (queryString) {
		nextPath += `?${queryString}`;
	}

	history.pushState(null, '', nextPath);
```

Lets ditch repetitions

src/js/index.js
```diff
@@ -96,14 +122,18 @@ window.addEventListener('load', () => {
 
 	if ('URLSearchParams' in window) {
 		search = new URLSearchParams(window.location.search);
+
 		if (search.has('tags')) {
 			selectr.setValue(search.get('tags').split(' '));
 		}
+
+		showExpired.checked = search.has('expired');
```
aslafy-z: If we go `expired=y`, this should be changed to `(search.get('expired') || null) === 'y'` or something.

src/js/index.js
```diff
@@ -83,6 +95,20 @@ function cascade(force = false) {
 	if (force || this === sortingInput) {
 		handleSort(this === sortingInput);
 	}
+
+	if (this === showExpired) {
```
aslafy-z: Any change this goes to a dedicated function (`handleExpired`) ?

src/js/index.js
```diff
@@ -83,6 +95,20 @@ function cascade(force = false) {
...
+		if (!search) {
+			search = new URLSearchParams(window.location.search);
```
aslafy-z: Is this necessary for this case only (`handleExpired`)?

data.json
```diff
@@ -146,7 +146,7 @@
   {
     "name": "Netlify",
     "difficulty": "easy",
-    "description": "Complete this <a href='https://swag.netlify.com/docs/'>survey</a> about using netlify for documentation and get stickers.",
+    "description": "Complete this <a href='https://swag.netlify.com/docs/'>survey</a> about using Netlify for documentation and get stickers.",
```
aslafy-z: This typo fix shouldn't be there.

data.json
```diff
@@ -47,11 +47,11 @@
   {
     "name": "Contentful",
     "difficulty": "easy",
-    "description": "Sign up from Contentful to get a free pair of socks to your door.",
+    "description": "Sign up to Contentful to get a free pair of socks to your door.",
```
aslafy-z: Another typo fix to move

data.json
aslafy-z: Moved to #248, and merged!

data.json
aslafy-z: Moved to #248, and merged!

src/js/index.js
vikaspotluri123: instead of 'y' or 'n', can we use '1' / '0'?

itaisteinherz: > instead of 'y' or 'n', can we use '1' / '0'?

In my opinion, that would make the URL unreadable and ambiguous. I think `y` and `n` are simple, succinct and easily understandable, and so we should keep using those.

src/js/index.js
itaisteinherz: Not really, but I'm not sure if this is even necessary at all. I put it there to be safe, but I guess we could just move it to `cascade`.

itaisteinherz:
> Just found out it was related to the Grammarly Chrome extension. Once I disabled it, it loads so fast!

This didn't seem to be the case for me. This may be because of Chrome since when trying to load the preview on a clean install of FireFox, everything seemed to work fine.

src/js/index.js
itaisteinherz: > If we go `expired=y`, this should be changed to `(search.get('expired') || null) === 'y'` or something.

Why do we need the `|| null` part?

aslafy-z:
@itaisteinherz it worked well on a private browsing session even on Chrome at my place. I think it may be due to another plugin you got, could you try to disable some to find the cause in your browser?

src/js/index.js
aslafy-z: It's not needed, get returns null if it doesn't find a matching value.

itaisteinherz:
@aslafy-z Just tested it out again and turns out my case matches yours - the preview loads very nicely when disabling Grammarly as well as on private browsing. Earlier on, I was just telling Grammarly to be off on the preview, which didn't really make any difference, but completely disabling the extension seemed to do the trick for me.

src/js/index.js
```diff
@@ -64,9 +72,12 @@ function handleTags() {
 	if (!search) {
```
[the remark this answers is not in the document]
itaisteinherz: I added it since the website might be used from a browser that doesn't support `URLSearchParams`, but I guess that in that case the website wouldn't work anyways.

src/js/index.js
```diff
@@ -11,10 +11,11 @@ const sort = {
 	DIFFICULTY_DESCENDING: (a, b) => a.dataset.difficulty < b.dataset.difficulty ? 1 : -1
 };
 
-const contentEl = document.querySelector('#content');
```
aslafy-z: These edits comes from the PR that updates `xo`, you need to revert these (or run `npm run lint:fix`)

itaisteinherz: Done ✓

src/js/index.js
```diff
@@ -14,6 +14,7 @@ const sort = {
 const contentEl = document.querySelector('#content');
 const filterInput = document.querySelector('#filter');
 const sortingInput = document.querySelector('#sorting');
+const showExpired = document.querySelector('#expired');
```
aslafy-z: Not so important but that line may go after the `tagSelect` one

src/js/index.js
```diff
@@ -77,13 +83,24 @@ function handleSort() {
...
+}
+
 // The cascade is the function which handles calling filtering and sorting swag
 function cascade(force = false) {
 	force |= handleDifficulty(this === filterInput);
 	force |= handleTags(Boolean(this.el));
 	if (force || this === sortingInput) {
 		handleSort(this === sortingInput);
 	}
+	handleExpired();
```
aslafy-z: lint reports one missing new line before this line:
```
 src/js/index.js:101:2
  ✖  101:2  Expected blank line before this statement.  padding-line-between-statements
```

itaisteinherz: For some reason I'm not seeing any of these linting errors when running `xo` locally 🤔

vikaspotluri123: Make sure you reinstall deps, the new version of XO added that rule

vikaspotluri123:
I'll try to review in a few hours

src/js/index.js
itaisteinherz: Thanks, that did the trick for me 👍🏻

data.json
```diff
@@ -24,7 +24,7 @@
     "reference": "https://community.aras.com/b/english/posts/join-aras-labs-hacktoberfest-2018",
     "image": "https://www.aras.com/-/media/Images/Content-Row-Images/im-an-innovator.png",
     "dateAdded": "2018-10-24T11:45:00.000Z",
-    "tags": ["clothing", "hacktoberfest", "stickers"]
+    "tags": ["clothing", "hacktoberfest", "expired", "stickers"]
```
vikaspotluri123: Can you mark Alexa as expired as well?

will close #252

src/js/index.js
```diff
@@ -46,28 +47,33 @@ function handleDifficulty(difficultyChanged) {
 	return false;
 }
 
-function handleTags() {
-	const tags = selectr.getValue();
+function updateUrl() {
```
vikaspotluri123: Since we already have a `handleTags` function, I think it might be better to add this in:

```js
for(let [key, value] of search) {
	if (!value.trim()) {
		search.delete(key);
	}
}
```

and drop the conditionals in the other parts of the file

src/js/index.js
```diff
@@ -77,13 +83,24 @@ function handleSort() {
 		.forEach(sortedChild => contentEl.append(sortedChild));
 }
 
+function handleExpired() {
+	if (showExpired.checked) {
+		search.set('expired', 'y');
```
vikaspotluri123: I think we should use booleans (i.e. `expired=true|false`) here, and handle it as part of `updateUrl`

vikaspotluri123: @aslafy-z I think using booleans for a boolean param would be better - what was your reasoning for using `y` or `n`?

Yash-Garg: Agree with you on that @vikaspotluri123 💯

src/js/index.js
```diff
@@ -46,28 +47,38 @@ function handleDifficulty(difficultyChanged) {
...
-	const tags = selectr.getValue();
+function updateUrl() {
+	let nextPath = window.location.pathname;
 
-	if (tags.length === 0) {
-		history.pushState(null, '', window.location.pathname);
-		return;
+	const tempParams = new URLSearchParams();
+	for (const [key, value] of search) {
+		if (value.trim()) {
```
vikaspotluri123: What's the point of using multiple instance of URLSearchParams?

If you use `if(!...) { search.delete(key) }` you would still get the same functionality without instantiating any new instances

itaisteinherz: I tried but that didn't work well. My theory is that deleting a param while iterating over `search` caused the loop to skip over the next param.

vikaspotluri123: Makes sense...

```js
const badParams = [];
for (...) {
	if (!value.trim()) {
		badParams.push(key);
	}
}

badParams.forEach(param => search.delete(value));
```

Should reduce complexity

src/js/index.js
```diff
@@ -77,13 +88,20 @@ function handleSort() {
 		.forEach(sortedChild => contentEl.append(sortedChild));
 }
 
+function handleExpired() {
```
vikaspotluri123: This function is only used in one place 🙂

vikaspotluri123:
@itaisteinherz found one last thing while we wait for @aslafy-z to respond!

src/js/index.js
vikaspotluri123: I think we should make handleExpired a part of handleTags because it's a special-case tag

src/js/index.js
aslafy-z: go for `expired=true`. See https://github.com/swapagarwal/swag-for-dev/pull/159#issuecomment-451221219 https://github.com/swapagarwal/swag-for-dev/pull/159#issuecomment-451371846, I proposed to ditch any value and @itaisteinherz proposed to use `y` or `n` instead.

itaisteinherz: @aslafy-z Sorry for blaming you 😅 I'll update this in a bit.

itaisteinherz: Done in 83ac6376f9fbc20b8756828a3eb5e702d6c27b29 👌🏻

src/js/index.js
```diff
@@ -46,44 +47,58 @@ function handleDifficulty(difficultyChanged) {
...
 
-	if (!search) {
-		return;
-	}
-
 	search.set('tags', tags.join(' '));
-	const newRelativePathQuery = `${window.location.pathname}?${search.toString()}`;
-	history.pushState(null, '', newRelativePathQuery);
+
+	search.set('expired', showExpired.checked ? 'true' : '');
```
vikaspotluri123:
```suggestion
	search.set('expired', showExpired.checked);
```

src/js/index.js
```diff
@@ -46,44 +47,58 @@ function handleDifficulty(difficultyChanged) {
...
+	}
+
+	history.pushState(null, '', nextPath);
 }
 
 // The cascade is the function which handles calling filtering and sorting swag
 function cascade(force = false) {
 	force |= handleDifficulty(this === filterInput);
-	force |= handleTags(Boolean(this.el));
 	if (force || this === sortingInput) {
```
vikaspotluri123: Just noticed this, what was the reasoning for switching the order that `handleTags` and `handleSort` are called?

src/js/index.js
```diff
@@ -97,14 +112,18 @@ window.addEventListener('load', () => {
 
 	if ('URLSearchParams' in window) {
 		search = new URLSearchParams(window.location.search);
+
 		if (search.has('tags')) {
 			selectr.setValue(search.get('tags').split(' '));
 		}
+
+		showExpired.checked = search.get('expired') === 'true';
```
vikaspotluri123:
```suggestion
		showExpired.checked = Boolean(search.get('expired'));
```

src/js/index.js
itaisteinherz: I did it since the logical and visual order in which (I think) we handle everything is:

1. Difficulty
2. Sorting
3. Tags
4. Expired

And so I wanted the code to reflect it.

src/js/index.js
itaisteinherz: I think @aslafy-z said we should remove the `expired` search param in case it is set to `false`, to keep the URL as short and simple as possible.
I think we could set `expired` to `false`, but not by default (only if the checkbox is deliberately deselected).

vikaspotluri123: It would get deleted by the updateUrl fn

`search.set('expired', false)`
`search.get('expired') => "false"`
`"false".trim() => "false"`
`!"false" => true`
  - Param is added to delete queue

itaisteinherz: Test

vikaspotluri123: I dun goofed!

```suggestion
	search.set('expired', showExpired.checked || '');
```

itaisteinherz:
@vikaspotluri123 @aslafy-z Ready to merge? 🙃

src/js/index.js
vikaspotluri123: I think there was a reason for putting tags before sorting but I don't remember right now

src/js/index.js
```diff
@@ -46,44 +47,58 @@ function handleDifficulty(difficultyChanged) {
...
 
-function handleSort() {
-	Array.from(contentEl.children)
-		.map(child => contentEl.removeChild(child))
-		.sort(sort[sortingInput.value])
-		.forEach(sortedChild => contentEl.append(sortedChild));
+function updateUrl() {
+	let nextPath = window.location.pathname;
+
+	const emptyParams = [];
```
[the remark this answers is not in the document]
vikaspotluri123: We can expand the variable names, but 👍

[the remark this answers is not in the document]
itaisteinherz: > This logic can be replaced by:

It can be, but doing so would make the code not very readable and maintainable (in my opinion, at least), and that's more important than having it more optimized as a one-liner. I think we should stick with the current `for...of` loop.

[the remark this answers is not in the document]
aslafy-z: WDYT of:

```javascript
Array.from(search).forEach(([key, value]) => {
	if (!value.trim()) {
		search.delete(key);
	}
});
```

It cut the needs for a intermediate array, I think it keeps code dry. If it still makes code less readable for you, we'd go your way.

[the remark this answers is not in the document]
vikaspotluri123: Either way is good with me 🎉 

Just putting this out there; this won't have an effect on performance by any means because we're dealing with literally 5 elements at most - it's just fun to think about!

> It cut the needs for a intermediate array

That's completely true, but it requires 2 iterations over the data because `Array.from()` does something similar to the `for...of` internally, and then we iterate over the returned array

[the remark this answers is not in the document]
itaisteinherz: > It cut the needs for a intermediate array, I think it keeps code dry. If it still makes code less readable for you, we'd go your way.

@aslafy-z I think we should keep the current implementation 🙃

> this won't have an effect on performance by any means because we're dealing with literally 5 elements at most

Also true 😅

itaisteinherz:
@aslafy-z @vikaspotluri123 Ping 🏓

aslafy-z:
I still have a concert about the `cascade` function. Is the force switch not needed for `handleTags`? @vikaspotluri123, would you mind checking that and then we can merge?

vikaspotluri123:
I think that's something we should look at separately. The main reason there's a check for the sort function is because it's pretty expensive to do - while the other filtering options only trigger visibility, the sort function has to clone the nodes and then sort them. Since the tags don't really do anything like that, it shouldn't be an issue

aslafy-z:
OK! Let's merge! Thanks a lot for your work @itaisteinherz!

aslafy-z:
Oops, just seen a issue with this PR: 
Try reloading this page with no-cache: https://deploy-preview-159--peaceful-chandrasekhar-efde8e.netlify.com/ (or https://devswag.io when deployed).
It initially shows the list including expired items and then exclude them when the `tags` selector has loaded.

@itaisteinherz @vikaspotluri123

itaisteinherz:
> It initially shows the list including expired items and then exclude them when the `tags` selector has loaded.

I noticed this as well when testing out the PR locally, but thought it was because of me running the website with the default development Gulp task. 

I'll look into the issues and try to push fixes ASAP.

itaisteinherz:
> Also: deploy-preview-159--peaceful-chandrasekhar-efde8e.netlify.com/?expired=false redirects to deploy-preview-159--peaceful-chandrasekhar-efde8e.netlify.com/?expired=true

The issue is caused by this line:
https://github.com/swapagarwal/swag-for-dev/blob/9f20c772e3c50bf6062c05464d8ca8c7a8ce2e89/src/js/index.js#L120

Since `Boolean('false') === true`, the user gets redirected.

PR #160, opened 2018-11-13
WIP: Release RSS, Atom and JSON feeds 

### LINKED ISSUE #142: Feature request: RSS/Atom feed
Loved your website and work.
But I would like to make one feature request.
Instead of email notifications, if you could add some sort of rss/atom feed to subscribe to, that will be awesome.

Cheers.

aslafy-z:

I built a npm module to enable creation of these files from the gulp pipeline using https://github.com/jpmonette/feed library.
See https://github.com/aslafy-z/gulp-feed.

Still need to set some values but I'd love to have your point on this.

Closes #142


vikaspotluri123:
Could you link to the feeds? I can't find it 👀

aslafy-z:
> Could you link to the feeds? I can't find it

Just updated OP with a link :wink:

vikaspotluri123:
> Just updated OP with a link 😉

Thanks 😄 are you currently working on integrating the feed into the site?

aslafy-z:
I'm not, feel free to take it up if you like to!

vikaspotluri123:
> I'm not, feel free to take it up if you like to!

I wish I had time 😔 my schedule is terrible until mid December!

Ghost (which I work with extensively) does a pretty good job generating SEO-related things, including RSS-type feeds.

Here's an example feed: https://blog.ghost.org/rss/

The tags are implemented using the `dc:category` node, and I think we should do the same. Additionally, do we need to add the `pubDate` node to all the items? I think it's needed for feed generators like IFTTT and Feedly to know how to order content, or what content is new 🤔

Also, I think we might want to use the image that's uploaded to netlify; if history is an issue, we can use the netlify commit preview url as the root

aslafy-z:
There is an open PR (https://github.com/jpmonette/feed/pull/64) about adding `category` field. We could use it once it has been merged (not sure it's still active, I'm gona take it up if they give no news). 

I'm pretty we'll mess lot of feed readers if our dates are all the sames, we need to find out how to keep the date. I think adding a `createdAt` (or similar) can do the trick.

I'm gonna update to use the generated images instead but I wanted to gatter feedback before. How whould history be an issue? What do you mean?

vikaspotluri123:
> How whould history be an issue? What do you mean?

For example, Gatsby's image changed, so we had to hotfix a replacement from an older build. If, for example, you use IFTTT to generate an email when a new feed item appears, it will hotlink to whatever you tell it to, which could be deleted in the future.

We can just use the `DEPLOY_URL` [env var](https://www.netlify.com/docs/continuous-deployment/#build-environment-variables)

swapagarwal:
Needs rebase.

swapagarwal:
The [feed](https://deploy-preview-160--peaceful-chandrasekhar-efde8e.netlify.com/assets/feed/rss.xml) has items sorted alphabetically. Let's use `dateAdded` to sort?

PR #161, opened 2018-11-14
Add added date for each swag item

aslafy-z:

Add added date in ISO format.

Needed for #160


vikaspotluri123:
Refs #160?

aslafy-z:
Nice! Need another approval before any other swag is added (please :D).

aslafy-z:
Does `dateAdded` tag fits you guys?

swapagarwal:
@aslafy-z How are we backfilling the data?

aslafy-z:
I used the issues / PR creation date as reference for existing items, you mean for next items?

swapagarwal:
The ideal way would be to use the merged timestamp in future, what do you think?

swapagarwal:
Thanks @aslafy-z! 🚀

PR #170, opened 2018-11-17
Short-circuit cachebust for development

### LINKED ISSUE #164: Errors while using gulp watch

`cachebust` task deletes `rev-manifest.json` while `gulp watch`ing.

**Steps to reproduce**:

- have a clean env: `gulp clean`
- run `gulp` and wait for build to finish
- edit any `src/js/*.js` `src/styl/**/*.styl` file and save
- the changes are picked up by gulp, but clean rule is called and so `rev-manifest.json` is deleted
- `cachebust` task can't recover image paths

**Output**:

```
[17:26:38] Starting 'clean'...
[17:26:38] Finished 'clean' after 18 ms
[17:26:38] Starting 'styl'...
[17:26:38] Finished 'styl' after 33 ms
[17:26:38] Starting 'cachebust'...
Unable to find image in manifest: swag-img/alexa-ae4c621281.jpg
Unable to find image in manifest: swag-img/codeship-44ee32c752.jpg
Unable to find image in manifest: swag-img/devrant-b5db1750ec.jpg
Unable to find image in manifest: swag-img/gatsby-e57212455c.jpg
Unable to find image in manifest: swag-img/google_assistant-0880ffbe3b.jpg
Unable to find image in manifest: swag-img/mullvad-960aef2001.jpg
Unable to find image in manifest: swag-img/netlify-7be19aa6f0.jpg
Unable to find image in manifest: swag-img/npm-16e8bc80de.jpg
Unable to find image in manifest: swag-img/passenger-2feca7e10f.jpg
Unable to find image in manifest: swag-img/whatismybrowser-96e64863b8.jpg
[17:26:38] Finished 'cachebust' after 17 ms
```

**`rev-manifest.json`**:

```json
{
  "css/index.css": "css/index-9159b10c46.css"
} 
```

How to solve this? Should we have a clean rule per file type?
Deleting the dependency to `clean` in each watch trigger does not work at all, some files are picked up multiple times if I do that:

```json
  "js/index-226706dba4.js": "js/index-226706dba4-226706dba4.js",
  "js/index.js": "js/index-226706dba4.js",
  "css/index-9159b10c46.css": "css/index-9159b10c46-9159b10c46.css",
  "css/index.css": "css/index-e18c66172d.css",
```

Any thoughts?

aslafy-z:
Fixes #164 

Implement suggestions from @vikaspotluri123 from https://github.com/swapagarwal/swag-for-dev/issues/164#issuecomment-439415961

- disable `cachebust` task for dev
- move `rev-manifest.json` to `dist/`

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,7 +1,7 @@
 [build]
   base      = "/"
   publish   = "dist"
-  command   = "gulp build"
+  command   = "npm run build"
 
 [build.environment]
   NODE_VERSION = "10"
--- a/.gitignore
+++ b/.gitignore
@@ -81,5 +81,3 @@ typings/
 
 
 # End of https://www.gitignore.io/api/node
-
-rev-manifest.json
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -14,6 +14,7 @@ const responsive    = require('gulp-responsive');
 
 const {swagList, swagImages} = require('./get-data');
 
+const REV_PATH = './dist/rev-manifest.json';
 const RESIZE_OPTS = {
     quality: 90,
     progressive: true,
@@ -35,6 +36,11 @@ gulp.task('pug', () => {
         },
         new Set()
     )).sort();
+
+    // import rev-manifest if it exists
+    try {
+        manifest = require(REV_PATH);
+    } catch(error) { }
 
     const bustedAssets = {
         css: `/assets/${manifest['css/index.css']}`,
@@ -109,7 +115,6 @@ gulp.task('clean:pug', () => del('dist/index.html'));
 gulp.task('clean', gulp.parallel('clean:pug', 'clean:assets'));
 
 gulp.task('cachebust', cb => {
-    const revPath = './dist/rev-manifest.json';
     const basePath = 'dist/assets';
     const bustedFiles = [
         'dist/assets/css/*',
@@ -122,13 +127,13 @@ gulp.task('cachebust', cb => {
         .pipe(cacheClean())
         .pipe(gulp.dest(basePath))
         .pipe(cachebust.manifest({
-            path: revPath,
+            path: REV_PATH,
             base: basePath,
         }))
         .pipe(gulp.dest(basePath))
         .on('end', () => {
-            delete require.cache[require.resolve(revPath)];
-            manifest = require(revPath);
+            delete require.cache[require.resolve(REV_PATH)];
+            manifest = require(REV_PATH);
 
             swagList.forEach(swag => {
                 const filename = `swag-img/${swag.image.split('/').pop()}`;
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -25,18 +25,10 @@ const RESIZE_OPTS = {
     errorOnUnusedConfig: false
 };
 
-let manifest;
-if (PRODUCTION) {
-    // import rev-manifest if it exists
-    try {
-        manifest = require(REV_PATH);
-    } catch(error) { }
-} else {
-    manifest = {
-        'css/index.css': 'css/index.css',
-        'js/index.js': 'js/index.js',
-    };
-}
+let manifest = {
+    'css/index.css': 'css/index.css',
+    'js/index.js': 'js/index.js',
+};
 
 gulp.task('pug', () => {
     const tags = Array.from(swagList.reduce(
@@ -46,6 +38,11 @@ gulp.task('pug', () => {
         },
         new Set()
     )).sort();
+
+    if (PRODUCTION) {
+        delete require.cache[require.resolve(REV_PATH)];
+        manifest = require(REV_PATH);
+    }
 
     const bustedAssets = {
         css: `/assets/${manifest['css/index.css']}`,
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -14,6 +14,11 @@ const responsive    = require('gulp-responsive');
 
 const {swagList, swagImages} = require('./get-data');
 
+let manifest = {
+    'css/index.css': 'css/index.css',
+    'js/index.js': 'js/index.js',
+};
+
 const RESIZE_OPTS = {
     quality: 90,
     progressive: true,
@@ -31,7 +36,6 @@ gulp.task('pug', () => {
         new Set()
     )).sort();
 
-    const manifest = require('./rev-manifest.json');
     const bustedAssets = {
         css: `/assets/${manifest['css/index.css']}`,
         js: `/assets/${manifest['js/index.js']}`
@@ -98,34 +102,34 @@ gulp.task('swag-img:optimize', () => {
 
 gulp.task('swag-img', gulp.series('swag-img:clean', 'swag-img:download' , 'swag-img:optimize'));
 
-gulp.task('clean', () => {
-    return del([
-        './rev-manifest.json',
-        'dist/assets/css/*',
-        'dist/assets/js/*',
-        'dist/assets/swag-img/*'
-    ]);
-});
+gulp.task('clean:styl', () => del('dist/assets/css/*'));
+gulp.task('clean:js', () => del('dist/assets/js/*'));
+gulp.task('clean:assets', gulp.parallel('clean:styl', 'clean:js'));
+gulp.task('clean:pug', () => del('dist/index.html'));
+gulp.task('clean:rev', () => del('dist/rev-manifest.json'));
+gulp.task('clean', gulp.parallel('clean:pug', 'clean:assets', 'clean:rev'));
 
 gulp.task('cachebust', cb => {
+    const revPath = './dist/rev-manifest.json';
+    const basePath = 'dist/assets';
     const bustedFiles = [
         'dist/assets/css/*',
         'dist/assets/js/*',
         'dist/assets/swag-img/*'
     ];
 
-    return gulp.src(bustedFiles, {base: 'dist/assets/'})
+    return gulp.src(bustedFiles, {base: basePath})
         .pipe(cachebust())
         .pipe(cacheClean())
-        .pipe(gulp.dest('dist/assets'))
-        .pipe(cachebust.manifest({
-            base: 'dist/assets',
-            merge: true
+        .pipe(gulp.dest(basePath))
+        .pipe(cachebust.manifest(revPath, {
+            base: basePath,
         }))
-        .pipe(gulp.dest('dist/assets/'))
+        .pipe(gulp.dest(basePath))
         .on('end', () => {
-            delete require.cache[require.resolve('./rev-manifest.json')];
-            const manifest = require('./rev-manifest.json');
+            delete require.cache[require.resolve(revPath)];
+            manifest = require(revPath);
+
             swagList.forEach(swag => {
                 const filename = `swag-img/${swag.image.split('/').pop()}`;
                 if (!manifest[filename]) {
@@ -148,9 +152,9 @@ gulp.task('webserver', () => {
 });
 
 gulp.task('watch', () => {
-    gulp.watch('src/pug/**/*.pug', gulp.parallel('pug'));
-    gulp.watch('src/styl/**/*.styl', gulp.series('clean', 'styl', 'cachebust'));
-    gulp.watch('src/js/*.js', gulp.series('clean', 'js', 'cachebust'));
+    gulp.watch('src/pug/**/*.pug', gulp.series('pug'));
+    gulp.watch('src/styl/**/*.styl', gulp.series('clean:styl', 'styl'));
+    gulp.watch('src/js/*.js', gulp.series('clean:js', 'js'));
 });
 
 gulp.task('build', gulp.series(
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -14,6 +14,8 @@ const responsive    = require('gulp-responsive');
 
 const {swagList, swagImages} = require('./get-data');
 
+const PRODUCTION = process.env.NODE_ENV === 'production';
+
 const REV_PATH = './dist/rev-manifest.json';
 const RESIZE_OPTS = {
     quality: 90,
@@ -23,10 +25,18 @@ const RESIZE_OPTS = {
     errorOnUnusedConfig: false
 };
 
-let manifest = {
-    'css/index.css': 'css/index.css',
-    'js/index.js': 'js/index.js',
-};
+let manifest;
+if (PRODUCTION) {
+    // import rev-manifest if it exists
+    try {
+        manifest = require(REV_PATH);
+    } catch(error) { }
+} else {
+    manifest = {
+        'css/index.css': 'css/index.css',
+        'js/index.js': 'js/index.js',
+    };
+}
 
 gulp.task('pug', () => {
     const tags = Array.from(swagList.reduce(
@@ -36,11 +46,6 @@ gulp.task('pug', () => {
         },
         new Set()
     )).sort();
-
-    // import rev-manifest if it exists
-    try {
-        manifest = require(REV_PATH);
-    } catch(error) { }
 
     const bustedAssets = {
         css: `/assets/${manifest['css/index.css']}`,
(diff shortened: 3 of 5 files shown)


site/gulpfile.js
```diff
@@ -1,18 +1,20 @@
...
-const pug           = require('gulp-pug');
-const htmlmin       = require('gulp-htmlmin');
-const stylus        = require('gulp-stylus');
-const webserver     = require('gulp-webserver');
-const concat        = require('gulp-concat');
-const download      = require('gulp-download-stream');
-const responsive    = require('gulp-responsive');
-
-const {swagList, swagImages} = require('./get-data');
+const del               = require('del');
```
vikaspotluri123: I think we need to just use one space to separate equal signs

```js
const package = require('package');
const superLongPackageName = require('superLongPackageName');
```

Otherwise, every time we modify our dependencies, the diffs will be unnecessarily bloated and I don't want to click extra buttons 😆

aslafy-z: I'm super ok with what you saying but as we discussed said in here (https://github.com/swapagarwal/swag-for-dev/pull/76#issuecomment-433082739) we should settle for coding conventions in another PR. I tried to stick with the current one in this PR so.

site/gulpfile.js
```diff
@@ -98,32 +102,34 @@ gulp.task('swag-img:optimize', () => {
...
-        .pipe(cachebust())
-        .pipe(cacheClean())
-        .pipe(gulp.dest('dist/assets'))
-        .pipe(cachebust.manifest({
-            base: 'dist/assets',
-            merge: true
-        }))
-        .pipe(gulp.dest('dist/assets/'))
+    return gulp.src(bustedFiles, {base: basePath})
+        .pipe(CACHEBUST ? cachebust() : noop())
```
vikaspotluri123: While this logic works, imo it's really unreadable. I think we should just do something like
```js
if (!CACHEBUST) {
	// glob is a subdep which we can hoist to the root, just use the same version :)
    const glob = require('glob');
	const {writeFileSync} = require('fs');
	// glob takes the glob as `{glob1,glob2,glob3}`
    const fileList = `{${bustedFiles.join(',')}}`;
	// this can be cleaned up a either using promises or glob.sync
    glob(fileList, (err, files) => {
		const json = files
			.map(file => file.replace('dist/assets/', ''))
			.reduce((state, file) => { state[file] = file; return state; }, {});
		fs.writeFileSync('filename', JSON.stringify(json, null, 2));
        cb();
    });
    return;
}

// Normal case
```

site/gulpfile.js
```diff
@@ -147,14 +153,21 @@ gulp.task('webserver', () => {
...
+});
+
 gulp.task('watch', () => {
-    gulp.watch('src/pug/**/*.pug', gulp.parallel('pug'));
-    gulp.watch('src/styl/**/*.styl', gulp.series('clean', 'styl', 'cachebust'));
-    gulp.watch('src/js/*.js', gulp.series('clean', 'js', 'cachebust'));
+    gulp.watch('../data.json', gulp.series('reload-swag', 'swag-img', 'cachebust', 'pug'));
+    gulp.watch('src/pug/**/*.pug', gulp.series('pug'));
+    gulp.watch('src/styl/**/*.styl', gulp.series('clean:styl', 'styl', 'cachebust'));
+    gulp.watch('src/js/*.js', gulp.series('clean:js', 'js', 'cachebust'));
```
vikaspotluri123: We can just run cachebust once at the beginning since the manifest won't change 🤔

site/gulpfile.js
vikaspotluri123: Of course I was the one that said something 😅 I meant this more as an observation rather than something that should be changed 👍

site/gulpfile.js
aslafy-z: What do you call the beginning? `swag-img`, `styl` and `js` tasks can change the manifest content, right?

vikaspotluri123: The manifest shouldn't change when watching because cachebusting should be disabled in development

aslafy-z: I dropped calls to cachebust for the watch `js` and `css` and it should do the trick as `cachebust` is ran by the `build` task so it has already ran when `watch` task starts. I keep if for watch `data` as it should react when adding / removing a swag item.

site/gulpfile.js
```diff
@@ -81,6 +87,8 @@ gulp.task('img', () => {
 });
 
 gulp.task('swag-img:download', () => {
+    ({ swagImages, swagList } = importFresh('./get-data'));
```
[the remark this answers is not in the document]
aslafy-z: Opened a PR to have a lint-rule that defines that. Let's discuss it #186!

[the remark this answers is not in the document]
aslafy-z: Eslint throws a syntax error if I throw the surrounding `()`.

site/get-data.js
```diff
@@ -1,5 +1,7 @@
 const swagImages = [];
 const fileNames = [];
+
+delete require.cache[require.resolve('../data.json')];
```
vikaspotluri123: If we're following the dev suggestion of not using `gulp watch` with changes to the `data.json` file, this isn't necessary

site/gulpfile.js
```diff
@@ -12,7 +13,15 @@ const concat        = require('gulp-concat');
 const download      = require('gulp-download-stream');
 const responsive    = require('gulp-responsive');
 
-const {swagList, swagImages} = require('./get-data');
+let swagList;
+let swagImages;
+
+const PRODUCTION = process.env.NODE_ENV === 'production';
+
+let manifest = {
```
vikaspotluri123: constantly const?

site/gulpfile.js
```diff
@@ -81,6 +89,9 @@ gulp.task('img', () => {
 });
 
 gulp.task('swag-img:download', () => {
+    delete require.cache[require.resolve('./get-data')];
```
vikaspotluri123: Shouldn't be needed (see comment in get-data)

site/gulpfile.js
```diff
@@ -98,34 +109,47 @@ gulp.task('swag-img:optimize', () => {
...
 gulp.task('cachebust', cb => {
+    const basePath = 'dist/assets/';
     const bustedFiles = [
         'dist/assets/css/*',
         'dist/assets/js/*',
         'dist/assets/swag-img/*'
     ];
 
-    return gulp.src(bustedFiles, {base: 'dist/assets/'})
+    if (!PRODUCTION) {
```
vikaspotluri123: This branch shouldn't be necessary - the only files read from the manifest are the scripts and styles, which are already in the dist folder

site/gulpfile.js
```diff
@@ -148,9 +167,10 @@ gulp.task('webserver', () => {
 });
 
 gulp.task('watch', () => {
-    gulp.watch('src/pug/**/*.pug', gulp.parallel('pug'));
-    gulp.watch('src/styl/**/*.styl', gulp.series('clean', 'styl', 'cachebust'));
-    gulp.watch('src/js/*.js', gulp.series('clean', 'js', 'cachebust'));
+    gulp.watch('../data.json', gulp.series('swag-img', 'cachebust', 'pug'));
```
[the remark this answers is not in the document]
aslafy-z: I haven't understood well the point you rose in the first time. I thought we could keep it as it does not add a big overhead to the code. It's pretty cool to see the website updating itself after I update an image or description field. You don't think?

site/gulpfile.js
aslafy-z: I currently reassign it in the `cachebust` task: https://github.com/swapagarwal/swag-for-dev/pull/170/files/15ea821f1a5da409a223cede8e5b7e439b0e2859#diff-6eb60389fb540a5c670ae4d43af865a9R131

aslafy-z:
Should be good this time!

site/.gitignore
```diff
@@ -81,5 +81,3 @@ typings/
 
 
 # End of https://www.gitignore.io/api/node
-
-rev-manifest.json
```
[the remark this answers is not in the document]
vikaspotluri123: Never mind, I misread it 😴

site/gulpfile.js
```diff
@@ -98,34 +102,34 @@ gulp.task('swag-img:optimize', () => {
...
     ];
 
-    return gulp.src(bustedFiles, {base: 'dist/assets/'})
+    return gulp.src(bustedFiles, {base: basePath})
         .pipe(cachebust())
         .pipe(cacheClean())
-        .pipe(gulp.dest('dist/assets'))
-        .pipe(cachebust.manifest({
-            base: 'dist/assets',
-            merge: true
```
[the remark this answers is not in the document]
aslafy-z: Diff shows a delete, but I just updated the paths.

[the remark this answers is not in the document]
aslafy-z: I think it's better to ask it to rewrite it every time. `merge` option needs to read the file first and adds unneeded logic in the background.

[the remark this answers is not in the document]
vikaspotluri123: I mean the `merge: true` option!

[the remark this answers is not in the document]
vikaspotluri123: Gotcha, carry on 🚶‍♂️

site/gulpfile.js
```diff
@@ -14,6 +14,11 @@ const responsive    = require('gulp-responsive');
 
 const {swagList, swagImages} = require('./get-data');
 
+let manifest = {
```
vikaspotluri123: Can we move this after the `RESIZE_OPTS` const? I like to follow the design pattern where you define constants first

site/gulpfile.js
```diff
@@ -98,34 +102,34 @@ gulp.task('swag-img:optimize', () => {
...
-        'dist/assets/css/*',
-        'dist/assets/js/*',
-        'dist/assets/swag-img/*'
-    ]);
-});
+gulp.task('clean:styl', () => del('dist/assets/css/*'));
+gulp.task('clean:js', () => del('dist/assets/js/*'));
+gulp.task('clean:assets', gulp.parallel('clean:styl', 'clean:js'));
+gulp.task('clean:pug', () => del('dist/index.html'));
+gulp.task('clean:rev', () => del('dist/rev-manifest.json'));
```
vikaspotluri123: Since we're overwriting the manifest, I think we can get rid of this. It's really only a scratch file

site/gulpfile.js
```diff
@@ -98,34 +102,34 @@ gulp.task('swag-img:optimize', () => {
...
-    return gulp.src(bustedFiles, {base: 'dist/assets/'})
+    return gulp.src(bustedFiles, {base: basePath})
         .pipe(cachebust())
         .pipe(cacheClean())
-        .pipe(gulp.dest('dist/assets'))
-        .pipe(cachebust.manifest({
-            base: 'dist/assets',
-            merge: true
+        .pipe(gulp.dest(basePath))
+        .pipe(cachebust.manifest(revPath, {
```
vikaspotluri123: This function should support passing the path in as part of the options object

`cache.manifest({path: revPath, /*...*/})`

netlify.toml
```diff
@@ -1,7 +1,7 @@
 [build]
   base      = "site/"
   publish   = "site/dist"
-  command   = "gulp build"
+  command   = "npm run build"
```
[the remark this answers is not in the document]
aslafy-z: I'd say `npm run` adds the local node_modules/.bin folder to PATH but netlify does it too

From https://www.netlify.com/docs/build-settings/:
> All of the installing of dependencies described here happens before your build command is run — so you can include webpack in your /package.json and be confident that it will be npm install‘ed before we run your build command webpack!

I think we have to use `npm scripts` for this kind of commands. If anyone try to reproduce `gulp build` in their term, it wont work out of the box, they will have to do either
- `npx gulp build`
- `./node_modules/.bin/gulp build`

but `npm run build` will work

site/gulpfile.js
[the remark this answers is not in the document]
aslafy-z: I don't see the point of keeping it tho

netlify.toml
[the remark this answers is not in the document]
vikaspotluri123: Step 3 in the development getting started guide is to install gulp globally 🤔 The reason I don't like using proxy methods in cases like this is it adds unnecessary bloat and, depending on your system, can hurt performance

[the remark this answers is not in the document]
aslafy-z: I reworked the doc in here: https://github.com/swapagarwal/swag-for-dev/pull/76/files#diff-d432571d5bbb8bd98a141819a9805ef5R77, still waiting for being rebased and merged.

Do you have such examples of performance issues using `npm run` proxy?

As netlify adds a path for this command to work I assume we can let it like this but I think we need to update the doc to use `npm run build`. We could still add a note about gulp cli.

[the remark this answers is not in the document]
vikaspotluri123: That's specifically for Docker, right? Since there are issues with integrating gulp + docker together?

I don't have examples, but from experience with WSL, the npm command would have to run, which includes running the npm-cli codebase, and then the gulp codebase needs to run (as it would be called from npm)

I just ran `time npm -v` in WSL and it took  ~4s, with `time gulp -v` taking ~8s

Running it again after files have been cached, the performance is better, with ~1.5s for both commands combined

site/gulpfile.js
```diff
@@ -31,7 +36,6 @@ gulp.task('pug', () => {
         new Set()
     )).sort();
 
-    const manifest = require('./rev-manifest.json');
```
vikaspotluri123: We should keep this to keep `gulp pug` fully functional

netlify.toml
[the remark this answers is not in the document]
aslafy-z: It's in the Docker PR but in the `Production/Local` section of the file. 
https://github.com/swapagarwal/swag-for-dev/blob/2815533d35d1bbe3681825b5ef7c9228e0f6bd73/site/README.md#local-1

Didn't know about this these lags, gonna commit this last change so we can merge this PR.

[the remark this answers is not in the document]
vikaspotluri123: This can definitely be a discussion we have in gitter 😛

site/gulpfile.js
```diff
@@ -22,6 +23,11 @@ const RESIZE_OPTS = {
     errorOnUnusedConfig: false
 };
 
+let manifest = {
```
vikaspotluri123: I think we need to determine this based on environment 🤔 we might run into an issue where the files aren't built but the manifest still exists 😵

aslafy-z: That's why I added the `clean:rev` in the first place. What do you think about keeping the said rule?

aslafy-z: It should be good by now

netlify.toml
```diff
@@ -3,6 +3,9 @@
   publish   = "site/dist"
   command   = "gulp build"
 
+[build.environment]
```
vikaspotluri123: Are you sure this doesn't happen by default?

vikaspotluri123: It does, but we need to move devDeps -> deps

aslafy-z: From https://www.netlify.com/docs/build-settings/:
> `NODE_ENV` is by default set to `development` (typical settings: production or development; setting globally allows all dependency installations to work with our caching.)

You are right, running `npm install` with `NODE_ENV` set to `production` should ignore dev dependencies. We can either use another var name, set it only for the build command (in the `command` field?) we move the devdeps to deps.

vikaspotluri123: Another option is to enable cachebusting within the file, although I think `npm run build` would be better

When running `gulp watch`, set the `SFD_CACHEBUST_DISABLED` to true and check for it

site/gulpfile.js
```diff
@@ -22,6 +25,19 @@ const RESIZE_OPTS = {
     errorOnUnusedConfig: false
 };
 
+let manifest;
+if (PRODUCTION) {
+    // import rev-manifest if it exists
+    try {
+        manifest = require(REV_PATH);
+    } catch(error) { }
+} else {
```
vikaspotluri123: Can we avoid the else branch by setting it as the default value? It's possible you're in production but requiring the manifest fails -> there is no manifest

aslafy-z: Yes! Done

aslafy-z:
Thank you guys! :fire:

PR #181, opened 2018-11-22
Fix Google Assistant promo image

itaisteinherz:

I've replaced the current Google Assistant promo image, which returns a 400 error when fetched, with a different image from [this article](https://medium.com/@amdcaruso/my-first-step-with-the-google-assistant-ab4c5ba04176).

### DIFF
--- a/data.json
+++ b/data.json
@@ -60,7 +60,7 @@
     "difficulty": "hard",
     "description": "Make an app for the Google Assistant available to users, get an exclusive Google Assistant t-shirt and a Google Home (if a certain number of users engage with it)!",
     "reference": "https://developers.google.com/actions/community/overview",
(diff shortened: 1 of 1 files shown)


swapagarwal:
The image needs to be center-cropped. 🤔

aslafy-z:
I opened #182 in the meanwhile, haven'. It addresses this center-crop problem by using a better-sized image that can do the trick.

itaisteinherz:
@swapagarwal Is this what you mean?

screen shot 2018-11-22 at 17 59 53

itaisteinherz:
@swapagarwal @aslafy-z This and #182 are pretty much identical, and I don't really mind which one gets merged. The important thing for one of them to get merged ASAP, in order to fix the build.

PR #186, opened 2018-11-26
Tweak lint: force spaces between brakets

aslafy-z:

Enable eslint `object-curly-spacing` rule to force having spaces between object curly braces.

This is a suggestion, I opened this to discuss the subject.

### DIFF
--- a/site/.eslintrc
+++ b/site/.eslintrc
@@ -26,6 +26,10 @@
       "error",
       "always"
     ],
-    "no-console": 0
+    "no-console": 0,
+    "object-curly-spacing": [
+      "error",
+      "always"
+    ]
   }
 }


vikaspotluri123:
Since code formatting is a hot topic, are there any issues with switching to a super opinionated one like [xo](https://github.com/xojs/xo)? I've used it and 99% of the rules are fine

aslafy-z:
`xo` looks good! I use [`standard`](https://github.com/standard/standard) often, it's also pretty good.
I think we really need to use one of these so we are good once and for all with code style.

vikaspotluri123:
I like XO partly because of [unicorn](https://github.com/sindresorhus/eslint-plugin-unicorn)

vikaspotluri123:
Gonna close this in favor of #189 - in the end, code style doesn't really make a difference, and since we're both in agreement for XO, we can continue with that!

PR #189, opened 2018-11-29
feat(lint) use xo linter

### LINKED ISSUE #188: Suggestion: Use a opinionated linter
## Issue regarding the website

Ref @vikaspotluri123 at https://github.com/swapagarwal/swag-for-dev/pull/186#issuecomment-441652111

I'd go for [`xo`](https://github.com/xojs/xo) with the `space: true` option enabled to use spaces instead of tabs, what do you think?

aslafy-z:

- Add `xo` linter
	* disabled rules: unicorn/prefer-spread (https://github.com/swapagarwal/swag-for-dev/pull/189#discussion_r237580197)
	* enable `browser` env for `src/js/**/*.js`
- Add auto lint on commit with `husky` and `lint-staged`.

Closes #188
Closes #186


site/src/js/index.js
```diff
@@ -1,103 +1,106 @@
-/* global Selectr */
+/* global window, document, Selectr */
```
vikaspotluri123: The environment should be specified in the XO config

site/package.json
```diff
@@ -30,12 +30,28 @@
...
-  "license": "ISC",
-  "bugs": {
-    "url": "https://github.com/swapagarwal/swag-for-dev/issues"
+  "scripts": {
+    "lint": "xo '**/*.js'",
+    "lint:fix": "xo --fix '**/*.js'",
+    "build": "gulp build"
+  },
+  "xo": {
+    "space": true
```
vikaspotluri123: Is there a reason for this override?

vikaspotluri123: Also, to expand on this... either way we should use 4 spaces to maintain legibility. With Promises and async, nested callbacks are virtually nonexistant which (afaik) was one of the big reasons for using 2 spaces

site/src/js/index.js
```diff
@@ -1,103 +1,106 @@
...
+  DIFFICULTY_ASCENDING: (a, b) => a.dataset.difficulty > b.dataset.difficulty ? 1 : -1,
+  DIFFICULTY_DESCENDING: (a, b) => a.dataset.difficulty < b.dataset.difficulty ? 1 : -1
 };
 
 const contentEl = document.getElementById('content');
 const filterInput = document.getElementById('filter');
 const sortingInput = document.getElementById('sorting');
 
-const activateElements = els => Array.from(els).forEach(node => node.classList.add(ACTIVE_CLASS));
+const activateElements = els => [...els].forEach(node => node.classList.add(ACTIVE_CLASS));
```
vikaspotluri123: I'd prefer if we don't spread for clientside code as transpilation adds a proxy function for this, which unnecessarily increases the bundle size

site/src/js/index.js
```diff
@@ -1,103 +1,106 @@
...
 const allowDifficultySelect = shouldAllow => sortingInput.querySelectorAll('.difficulty')
-    .forEach(node => node.disabled = !shouldAllow);
+  .forEach(node => {
```
vikaspotluri123: Was there a specific reason for multi-lining this? It felt legible on a single line 😉

site/src/js/index.js
aslafy-z: it's from the default unicorn rules: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/prefer-spread.md

site/src/js/index.js
aslafy-z: This is also default with xo: https://github.com/xojs/eslint-config-xo/blob/master/index.js#L116

vikaspotluri123: Aah, forgot about that 🤮 could you try adding braces around it? I don't remember if that would trigger a `one-statement-per-line`-like error

`.foreach(node => {node.disabled = !shouldAllow})`

site/src/js/index.js
vikaspotluri123: Yeah, in this case I think it would be better to override this rule for the client js 🤔

aslafy-z:
https://github.com/xojs/xo/pull/279 prevents the use of `lint-staged` when config overrides are in `package.json`

site/src/js/index.js
aslafy-z: As of https://github.com/xojs/xo/pull/279 I think we can keep these globals in this file from now.

Here's the config for reference
```
  "xo": {
    "overrides": [
      {
        "files": "src/js/**/*.js",
        "globals": ["document", "window"]
      }
    ]
  },
```

vikaspotluri123: Instead of `globals` you can use `envs: ["browser"]`

site/src/js/index.js
aslafy-z: Yes it triggers it, that's why I multi-lined it :/

vikaspotluri123: Oh well, you win some you lose some! 😄

site/package.json
aslafy-z: moved to 4 spaces

site/src/js/index.js
aslafy-z: rule disabled with `eslint-disable`

Here's the config for reference
```
  "xo": {
    "overrides": [
      {
        "files": "src/js/**/*.js",
        "rules": {
          "unicorn/prefer-spread": "off"
        }
      }
    ]
  },
```

site/package.json
```diff
@@ -30,12 +30,44 @@
...
+    "husky": "^1.2.0",
+    "lint-staged": "^8.1.0",
+    "xo": "github:aslafy-z/xo#use-cosmiconfig"
   },
-  "author": "Swapnil Agarwal",
-  "license": "ISC",
-  "bugs": {
-    "url": "https://github.com/swapagarwal/swag-for-dev/issues"
+  "scripts": {
+    "lint": "xo '**/*.js'",
```
itaisteinherz:
```suggestion
    "lint": "xo",
```

XO already lints all JS files in all directories by default.

site/package.json
```diff
@@ -30,12 +30,44 @@
...
+    "lint:fix": "xo --fix '**/*.js'",
```
itaisteinherz:
```suggestion
    "lint:fix": "xo --fix",
```

Same as above.

site/package.json
```diff
@@ -30,12 +30,44 @@
...
+    "build": "gulp build"
+  },
+  "xo": {
+    "space": 4,
+    "plugins": [
+      "unicorn"
```
itaisteinherz: XO already [includes the unicorn plugin by default](https://github.com/xojs/xo#highlights), and so there's no need to manually include it.

aslafy-z: It's needed as I disable a unicorn rules in the rules. Custom behavior like this one need a implicit import of the said plugin. Don't find the reference to it just now but it doesn't work without.

itaisteinherz: From my testing of this branch locally without the `plugins` property specified in XO's config, linting seems to work properly (and the `unicorn/prefer-spread` rule stays disabled).
Is there anything specific that doesn't work for you when the `plugins` property isn't specified?

swapagarwal:
Thanks @itaisteinherz for chiming in to review! ❤️
@vikaspotluri123 Let's merge if it looks good?

vikaspotluri123:
Needs a rebase, I'll do a final review afterwards 😊

site/package.json
aslafy-z: Well it works as you said, I made the change!

site/package.json
```diff
@@ -30,12 +30,43 @@
     "gulp-rev-delete-original": "^0.2.3",
     "gulp-stylus": "^2.7.0",
     "gulp-uglify-es": "^1.0.4",
-    "gulp-webserver": "^0.9.1"
+    "gulp-webserver": "^0.9.1",
+    "husky": "^1.2.0",
+    "lint-staged": "^8.1.0",
+    "xo": "github:aslafy-z/xo#fix/lint-staged-absolute"
```
aslafy-z: Note that I'm using a fork, see here: https://github.com/xojs/xo/pull/355 https://github.com/xojs/xo/pull/279

aslafy-z: https://github.com/swapagarwal/swag-for-dev/pull/169 would solve it as we would be able to use the `relative` option of `lint-staged`.

vikaspotluri123: Does that mean you want to wait for that PR?

aslafy-z: It would be easier, or we can merge using my fork until we change the website root.
@itaisteinherz are you willing to finish it soon? I can rebase your work so we can go through if you want

itaisteinherz: I’m pretty sure that #169 is ready to be merged 👍🏻
I’ll be away from my computer for the next 3 days, so if you have anything you’d like to change feel free to push that to the PR. I can also talk on GitHub / Gitter if you’d like me to give feedback on anything.

aslafy-z: Nice! I've updated #169, it needed a final rebase.
This is going through! :tada:

vikaspotluri123:
@aslafy-z what PR needs to be merged before this one?

aslafy-z:
#169, see https://github.com/swapagarwal/swag-for-dev/pull/189#discussion_r238252959

aslafy-z:
I restored the original `xo` version as it works after #169 as planned! 
This is ready to merge!

package.json
```diff
@@ -30,12 +30,43 @@
...
+  "scripts": {
+    "lint": "xo",
+    "lint:fix": "xo --fix",
+    "build": "gulp build"
+  },
+  "xo": {
+    "rules": {
+      "unicorn/prefer-spread": "off"
+    },
+    "overrides": [
```
vikaspotluri123: Maybe we should do an inverse override, since only the gulpfile is not browser specific - default env is browser, and gulpfile is configured to be node

swapagarwal:
@aslafy-z lgtm, feel free to merge!

vikaspotluri123:
Of course I remembered after I clicked the button!

Do we want to [badge it up](https://github.com/xojs/xo#badge)?

swapagarwal:
Sure, I think small additions can be directly made to master. Or do you want everything to go through our review process?

vikaspotluri123:
Small additions don't need to PR'd, we just need to make sure it won't trigger the need for rebases everywhere 🙈

vikaspotluri123:
With this specifically, it should be fine. I'm just saying if you're going to commit directly to master, be careful!

But also, for things like this I don't think we need review approvals?

swapagarwal:
Cool, added! 🎉

PR #190, opened 2018-11-29
chore(licence) update package.json licence field

aslafy-z:

Update `package.json` licence field to match [LICENSE file](https://github.com/swapagarwal/swag-for-dev/blob/master/LICENSE).


swapagarwal:
Thanks @aslafy-z! 👍

PR #197, opened 2018-12-04
feat: (Closes #183) Implement "difficulty" and "sort" handling from URL params 

### LINKED ISSUE #183: Handle Difficulty & Ordering from URL

Handle updates for

- difficulty
- sorting
 
from URL. This was we can link to any sort of settings (eg: https://degswag.io/?tags=stickers&difficulty=easy&sorting=alphabetical_asc)

plibither8:
This PR implements #183: something which I too was wanting to suggest for quite some time.

I have coded this in quite a hurry so it might seem sloppy. There is more work to be put into this PR, IMHO.

*Summary*:

* Created a main `parameters` Object that holds `action`, `getValue` and `setValue` methods for the parameters specified in the Object. Currently, they are: `tags`, `difficulty` and `sort`.

* New function `updateUrl` which is called in the `handleTags` function. It updates the URL whenever there is a change in the input of either of the three parameters.

*NB: I might not be available that often to reply or maintain this PR and other related issues due to exams. Please feel free to directly commit to this PR when need be :smile:*

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -19,10 +19,53 @@ const activateElements = els => Array.from(els).forEach(node => node.classList.a
 const allowDifficultySelect = shouldAllow => sortingInput.querySelectorAll('.difficulty')
     .forEach(node => node.disabled = !shouldAllow);
 
+const parameters = {
+    'tags': {
+        action: value => selectr.setValue(value.split(' ')),
+        getValue: () => selectr.getValue().join(' '),
+        setValue: value => { return; }
+    },
+    'difficulty': {
+        action: value => {
+            switch (value.toLowerCase()) {
+                case 'easy':
+                case 'medium':
+                case 'hard':
+                    handleDifficulty(value.toLowerCase());
+                default:
+                    handleDifficulty();
+            }
+        },
+        getValue: () => filterInput.value,
+        setValue: value => filterInput.value = value
+    },
+    'sort': {
+        action: value => {
+            if (value in sort) {
+                handleSort(sort[value]);
+            }
+        },
+        getValue: () => sortingInput.value,
+        setValue: value => sortingInput.value = value
+    }
+};
+
 let search, selectr;
 
-function handleDifficulty(difficultyChanged) {
-    const {value} = filterInput;
+function updateUrl() {
+    let newSearch = new URLSearchParams(window.location.search);
+    for (let parameter in parameters) {
+        if (parameters[parameter].getValue() === '') {
+            newSearch.delete(parameter);
+            continue;
+        }
+        newSearch.set(parameter, parameters[parameter].getValue());
+    };
+    const newRelativePathQuery = `${window.location.pathname}?${newSearch.toString()}`;
+    history.pushState(null, '', newRelativePathQuery);
+}
+
+function handleDifficulty(difficultyChanged = false, value = 'alldifficulties') {
     Array.from(contentEl.getElementsByClassName(ACTIVE_CLASS))
         .forEach(swag => swag.classList.remove(ACTIVE_CLASS));
 
@@ -46,7 +89,7 @@ function handleTags() {
     const tags = selectr.getValue();
 
     if (!tags.length) {
-        history.pushState(null, '', window.location.pathname);
+        updateUrl();
         return;
     }
 
@@ -60,24 +103,23 @@ function handleTags() {
     if (!search) {
         return;
     }
-    search.set('tags', tags.join(' '));
-    const newRelativePathQuery = `${window.location.pathname}?${search.toString()}`;
-    history.pushState(null, '', newRelativePathQuery);
+
+    updateUrl();
 }
 
-function handleSort() {
+function handleSort(sortValue) {
     Array.from(contentEl.children)
         .map(child => contentEl.removeChild(child))
-        .sort(sort[sortingInput.value])
+        .sort(sortValue)
         .forEach(sortedChild => contentEl.appendChild(sortedChild));
 }
 
 // The cascade is the function which handles calling filtering and sorting swag
 function cascade(force = false) {
-    force |= handleDifficulty(this === filterInput);
+    force |= handleDifficulty(this === filterInput, filterInput.value);
     force |= handleTags(Boolean(this.el));
     if (force || this === sortingInput) {
-        handleSort(this === sortingInput);
+        handleSort(sort[sortingInput.value]);
     }
 }
 
@@ -90,8 +132,13 @@ window.addEventListener('load', () => {
 
     if ('URLSearchParams' in window) {
         search = new URLSearchParams(window.location.search);
-        if (search.has('tags')) {
-            selectr.setValue(search.get('tags').split(' '));
+
+        for (let parameter in parameters) {
+            if (search.has(parameter)) {
+                let parameterValue = search.get(parameter);
+                parameters[parameter].action(parameterValue);
+                parameters[parameter].setValue(parameterValue);
+            }
         }
     }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -3426,13 +3426,15 @@
           "version": "1.0.0",
           "resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-1.0.0.tgz",
           "integrity": "sha1-ibTRmasr7kneFk6gK4nORi1xt2c=",
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.11.tgz",
           "integrity": "sha512-iCuPHDFgrHX7H2vEI/5xpz07zSHB00TpugqhmYtVmMO6518mCuRMoOYFldEBl0g187ufozdaHgWKcYFb61qGiA==",
           "dev": true,
+          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
@@ -3449,19 +3451,22 @@
           "version": "1.1.0",
           "resolved": "https://registry.npmjs.org/code-point-at/-/code-point-at-1.1.0.tgz",
           "integrity": "sha1-DQcLTQQ6W+ozovGkDi7bPZpMz3c=",
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "concat-map": {
           "version": "0.0.1",
           "resolved": "https://registry.npmjs.org/concat-map/-/concat-map-0.0.1.tgz",
           "integrity": "sha1-2Klr13/Wjfd5OnMDajug1UBdR3s=",
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "console-control-strings": {
           "version": "1.1.0",
           "resolved": "https://registry.npmjs.org/console-control-strings/-/console-control-strings-1.1.0.tgz",
           "integrity": "sha1-PXz0Rk22RG6mRL9LOVB/mFEAjo4=",
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "core-util-is": {
           "version": "1.0.2",
@@ -3592,7 +3597,8 @@
           "version": "2.0.3",
           "resolved": "https://registry.npmjs.org/inherits/-/inherits-2.0.3.tgz",
           "integrity": "sha1-Yzwsg+PaQqUC9SRmAiSA9CCCYd4=",
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "ini": {
           "version": "1.3.5",
@@ -3606,6 +3612,7 @@
           "resolved": "https://registry.npmjs.org/is-fullwidth-code-point/-/is-fullwidth-code-point-1.0.0.tgz",
           "integrity": "sha1-754xOG8DGn8NZDr4L95QxFfvAMs=",
           "dev": true,
+          "optional": true,
           "requires": {
             "number-is-nan": "^1.0.0"
           }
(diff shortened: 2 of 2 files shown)


plibither8:
### Note: the latest push to the PR is NOT working

The latest commit is my attempt at a neater implementation of the same, though it is not working the way it should. URL is updating properly but somehow (in a way I am not being able to grasp), all items are becoming in-`.visible`.

@vikaspotluri123 since you have worked on the cascade function, could you provide some insights? Thanks :smile:.

plibither8:
@vikaspotluri123 Ah yes! Should've caught that earlier :flushed:!

Nevertheless, the issue persists :slightly_frowning_face:

plibither8:
Problem fixed! :tada: Turns out I was using an outdated version of the JS file :see_no_evil: Don't cachebust during development! :smile:

vikaspotluri123:
Hopefully getting #170 merged will clean up your issues with cachebusting 🤞

aslafy-z:
I made a patch to support cleaning values if they are set to the default one, and not adding `?` in url if there are no args.

> Note: `pushState` does not create a history entry if the url is same as the current one

> Note: issue regarding URLSearchParams size check: https://github.com/whatwg/url/issues/163

```diff
diff --git a/src/js/index.js b/src/js/index.js
index 03c7499..e16c931 100644
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -21,14 +21,17 @@ const allowDifficultySelect = shouldAllow => sortingInput.querySelectorAll('.dif
 
 const parameters = {
     'tags': {
+        default: '',
         setValue: value => selectr.setValue(value.split(' ')),
         getValue: () => selectr.getValue().join(' ')
     },
     'difficulty': {
+        default: 'alldifficulties',
         setValue: value => filterInput.value = ['easy', 'medium', 'difficult'].includes(value) ? value : 'alldifficulties',
         getValue: () => filterInput.value,
     },
     'sort': {
+        default: 'ALPHABETICAL_ASCENDING',
         setValue: value => sortingInput.value = value in sort ? value : 'ALPHABETICAL_ASCENDING',
         getValue: () => sortingInput.value
     }
@@ -40,13 +43,17 @@ function updateUrl() {
     let newSearch = new URLSearchParams(window.location.search);
     for (let parameter in parameters) {
         const paramValue = parameters[parameter].getValue();
-        if (paramValue === '') {
+        if (['', parameters[parameter].default].includes(paramValue)) {
             newSearch.delete(parameter);
             continue;
         }
         newSearch.set(parameter, paramValue);
     }
-    const newRelativePathQuery = `${window.location.pathname}?${newSearch.toString()}`;
+    const newSearchString = newSearch.toString();
+    let newRelativePathQuery = window.location.pathname;
+    if (newSearchString) {
+        newRelativePathQuery += `?${newSearchString}`
+    }
     history.pushState(null, '', newRelativePathQuery);
 }
```

plibither8:
Alright! Will make the changes. Thanks :+1:

src/js/index.js
```diff
@@ -21,9 +21,51 @@ const allowDifficultySelect = shouldAllow => sortingInput.querySelectorAll('.dif
...
 	});
 
+const parameters = {
+	tags: {
+		default: '',
+		getValue: () => selectr.getValue().join(' '),
+		setValue: value => selectr.setValue(value.split(' '))
+	},
+	difficulty: {
+		default: 'alldifficulties',
```
vikaspotluri123: We should rename this from `alldifficulties` to `all`

src/js/index.js
```diff
@@ -21,9 +21,51 @@ const allowDifficultySelect = shouldAllow => sortingInput.querySelectorAll('.dif
...
+		getValue: () => filterInput.value,
+		setValue: value => {
+			filterInput.value = ['easy', 'medium', 'difficult'].includes(value) ? value : 'alldifficulties';
+		}
+	},
+	sort: {
+		default: 'ALPHABETICAL_ASCENDING',
```
vikaspotluri123: we should split this into `order` and `sortMethod`

i.e. `?sortMethod=alphabetical&order=asc`

src/js/index.js
```diff
@@ -80,8 +114,9 @@ function cascade(force = false) {
 	force |= handleDifficulty(this === filterInput);
 	force |= handleTags(Boolean(this.el));
 	if (force || this === sortingInput) {
-		handleSort(this === sortingInput);
+		handleSort();
```
vikaspotluri123: Why did we get rid of the boolean here?

plibither8:
@aslafy-z made the respective changes, please check :smile:

src/js/index.js
```diff
@@ -21,15 +21,74 @@ const allowDifficultySelect = shouldAllow => sortingInput.querySelectorAll('.dif
...
 let search;
 let selectr;
+const sortParams = {
+	sortMethod: sortingInput.value.split('_')[0],
+	order: sortingInput.value.split('_')[1]
+};
+
+function updateUrl() {
+	const newSearch = new URLSearchParams(window.location.search);
+	for (const parameter in parameters) {
```
aslafy-z: If you want to stick with a `for` loop, you want to use [`for ... of {}`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) instead of `for ... in {}` you uses so you can ditch that ugly hasOwnProperty check.

src/js/index.js
```diff
@@ -21,15 +21,74 @@ const allowDifficultySelect = shouldAllow => sortingInput.querySelectorAll('.dif
...
+		if (Object.prototype.hasOwnProperty.call(parameters, parameter)) {
+			const paramValue = parameters[parameter].getValue();
+			if (paramValue === '') {
```
aslafy-z: It misses the part I added to remote values that are set to default from the SearchString.

```suggestion
			if (['', parameters[parameter].default].includes(paramValue)) {
```

src/js/index.js
```diff
@@ -21,15 +21,74 @@ const allowDifficultySelect = shouldAllow => sortingInput.querySelectorAll('.dif
...
+	history.pushState(null, '', newRelativePathQuery);
+}
 
 function handleDifficulty(difficultyChanged) {
 	const {value} = filterInput;
 	Array.from(contentEl.getElementsByClassName(ACTIVE_CLASS))
 		.forEach(swag => swag.classList.remove(ACTIVE_CLASS));
 
-	if (value === 'alldifficulties') {
+	if (value === 'all') {
```
aslafy-z: We can use the default value we store in here.

```suggestion
	if (value === parameters.difficulty.default) {
```

src/js/index.js
```diff
@@ -49,7 +108,6 @@ function handleTags() {
 	const tags = selectr.getValue();
```
aslafy-z: **EDIT: won't work as the list is joined.**

Here also, let's use the getter you wrote.

```suggestion
	const tags = parameters.tags.getValue();
```

src/js/index.js
```diff
@@ -21,15 +21,74 @@ const allowDifficultySelect = shouldAllow => sortingInput.querySelectorAll('.dif
...
+				'descending' :
+				'ascending';
+		}
+	}
+};
+
 let search;
 let selectr;
+const sortParams = {
+	sortMethod: sortingInput.value.split('_')[0],
```
aslafy-z: Let's use `sort` everywhere instead of `sortMethod`.

aslafy-z: These values are set by the handleSort function, lets set them to empty, falsy value.

src/js/index.js
```diff
@@ -93,14 +149,16 @@ window.addEventListener('load', () => {
 
 	if ('URLSearchParams' in window) {
 		search = new URLSearchParams(window.location.search);
-		if (search.has('tags')) {
-			selectr.setValue(search.get('tags').split(' '));
+		for (const parameter in parameters) {
```
aslafy-z: This `for` loop lacks hasOwnProperty check, see [my other comment](https://github.com/swapagarwal/swag-for-dev/pull/197/files#diff-99a85c3932c0670278158f38d818e0efR68).

plibither8:
@aslafy-z `for...of` loops cannot be used to iterate over objects: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of.

Is there another solution?

plibither8:
Also, now with the URL updating _only_ when the params don't match default values, it somehow seems incomplete. For example, if I want to see the items in "Alphabetical, reversed" order, the URL is: https://devswag.io/?order=desc. What do you think about it?

plibither8:
Pinging @aslafy-z, @vikaspotluri123 for reviewal 😄

src/js/index.js
```diff
@@ -21,15 +21,73 @@ const allowDifficultySelect = shouldAllow => sortingInput.querySelectorAll('.dif
...
+			sortParams.sort = ['alphabetical', 'difficulty'].includes(value) ?
+				value.toUpperCase() :
+				'ALPHABETICAL';
+		}
+	},
+	order: {
+		default: 'asc',
+		getValue: () => `${sortParams.order.split('SC')[0]}sc`.toLowerCase(),
+		setValue: value => {
+			sortParams.order = value === 'desc' ?
```
vikaspotluri123: Since these values are internal, we can just rename them

i.e. `option[value=alpha_asc]` rather than `option[value=ALPHABETICAL_ASCENDING]`

src/js/index.js
```diff
@@ -59,28 +116,31 @@ function handleTags() {
...
+	sortingInput.value = Object.values(sortParams).join('_');
 	Array.from(contentEl.children)
 		.map(child => contentEl.removeChild(child))
 		.sort(sort[sortingInput.value])
 		.forEach(sortedChild => contentEl.appendChild(sortedChild));
 }
 
 // The cascade is the function which handles calling filtering and sorting swag
-function cascade(force = false) {
+function cascade(force = false, firstRun = false) {
```
vikaspotluri123: Force should just be set to true on firstRun, that's basically what it's designed for

src/pug/includes/body.pug
```diff
@@ -18,7 +18,7 @@ include ./fork.pug
                         label.sr-only(for="filter") Filter opportunity difficulty
                         select#filter
                             for filter in ['All difficulties', 'Easy', 'Medium', 'Hard']
```
vikaspotluri123: We can just update the array value rather than adding the string splitting logic

src/pug/includes/body.pug
```diff
@@ -18,7 +18,7 @@ include ./fork.pug
...
-                                option(value=filter.replace(/ /g, '').toLowerCase()) #{filter}
+                                option(value=filter.split(' ')[0].toLowerCase()) #{filter}
```
vikaspotluri123: Sorry if I wasn't clear, I was suggesting the UI be `select[all, easy, medium, hard] difficulties`

so when you chose medium it would effectively say `medium difficulties` rather than just medium which loses context

aslafy-z: I think we should `difficulties` can be moved outside the select

plibither8:
@aslafy-z moved 'difficulties' out of the select element 👍

src/js/index.js
```diff
@@ -5,10 +5,10 @@
  */
 const ACTIVE_CLASS = 'visible';
 const sort = {
-	ALPHABETICAL_ASCENDING: (a, b) => a.dataset.name > b.dataset.name ? 1 : -1,
-	ALPHABETICAL_DESCENDING: (a, b) => a.dataset.name < b.dataset.name ? 1 : -1,
-	DIFFICULTY_ASCENDING: (a, b) => a.dataset.difficulty > b.dataset.difficulty ? 1 : -1,
-	DIFFICULTY_DESCENDING: (a, b) => a.dataset.difficulty < b.dataset.difficulty ? 1 : -1
+	ALPHABETICAL_ASC: (a, b) => a.dataset.name > b.dataset.name ? 1 : -1,
```
vikaspotluri123: Does this need to be uppercase?

src/pug/includes/body.pug
```diff
@@ -17,18 +17,19 @@ include ./fork.pug
                     .select-wrapper
                         label.sr-only(for="filter") Filter opportunity difficulty
                         select#filter
-                            for filter in ['All difficulties', 'Easy', 'Medium', 'Hard']
-                                option(value=filter.replace(/ /g, '').toLowerCase()) #{filter}
+                            for filter in ['All', 'Easy', 'Medium', 'Hard']
```
vikaspotluri123: I think we should lowercase this as well as the sort-order input to make it more like a choose-your-own journey

src/js/index.js
```diff
@@ -21,15 +21,73 @@ const allowDifficultySelect = shouldAllow => sortingInput.querySelectorAll('.dif
...
+};
+
 let search;
 let selectr;
 
+function updateUrl() {
+	const newSearch = new URLSearchParams(window.location.search);
+	for (const [paramName, paramObj] of Object.entries(parameters)) {
+		const paramValue = paramObj.getValue();
+		if (['', paramObj.default].includes(paramValue)) {
```
itaisteinherz:
```suggestion
		if (paramValue.length === 0 || paramValue === paramObj.default) {
```

Makes the code more readable and clear, in my opinion.

aslafy-z:
@plibither8 Hello! Any chance you rebase your work? Thank you!

aslafy-z:
@plibither8 quick ping. Can you rebase this PR on last changes? Thanks!

aslafy-z:
@plibither8 We'd love to see this land, any chance you get some time to rebase the PR? Thank you!

plibither8:
@aslafy-z Hi! First of all sorry, I did not realise i completely ignored the previous two pings that you did, should've realised and replied earlier. I'm quite busy these days with other commitments but would be happy to give it a shot once I get time :). I''ll get back to you and update this thread soon, hopefully.

plibither8:
@aslafy-z in general, I think this site could go through a complete refactor to cleanup the code and make it more modern and maintainable. I started work on the site almost three years ago when I was comfortable working with Pug and Gulp.

@swapagarwal @aslafy-z and rest of the maintainers, if you have the capacity we can discuss a revival of Swag for Dev, since the maintenance and content update frequency have fell rather flat. Let's see :D.

PR #72, opened 2018-10-03
Website docs 

Jacob8765:
Adds a `README.md` file with some basic instructions on how to start the webserver. Fixes #71

PR #202, opened 2018-12-06
HTTPS links

coliff:
HTTPS links to site

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Sign up at [https://upscri.be/67d2fa](https://upscri.be/67d2fa) to get notified 
 
 ## Contents
 
-All swag opportunities can be found at [http://devswag.io/](http://devswag.io/) 😎
+All swag opportunities can be found at [https://devswag.io/](https://devswag.io/) 😎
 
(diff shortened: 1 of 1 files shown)


vikaspotluri123:
relevant thread: https://github.com/swapagarwal/swag-for-dev/pull/72#discussion_r222594121

aslafy-z:
I'd like to discuss more about this.

I don't see the point to use HTTPs in the first time if all the links we distribute are HTTP ones. The whole point of HTTPs is preventing mitm attacks, but is users are comming from an HTTP link, their are vulnerable.

If someone hijack the connection before the user is redirected, he'll be able to do whatever he wants with the user: giving him some poisoned version of devswag so he won't notice, or redirect to their own url or something. 

This is partly mitigated by using HSTS (Strict-Transport-Security header in `netlify.toml`) but still, I don't see the point. 

@swapagarwal you where talking about letting two doors open instead of one, if we don't offer a plain http version of the website it doesn't make sense from my point of view.

swapagarwal:
Agreed, this can be merged.

PR #205, opened 2018-12-06
Fix: Update HSTS policies to match listing requirements

aslafy-z:
This PR fixes some issues who prevent to be listed in the [hstspreload list](https://hstspreload.org):

- Error: No includeSubDomains directive
  The header must contain the `includeSubDomains` directive.
- Error: Max-age too low
  The max-age must be at least 31536000 seconds (≈ 1 year), but the header currently only has max-age=15552000.

See here: https://hstspreload.org/?domain=devswag.io

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -18,7 +18,7 @@
     Referrer-Policy = "strict-origin-when-cross-origin"
     X-XSS-Protection = "1; mode=block"
     Feature-Policy = "accelerometer 'none'; camera 'none'; geolocation 'none'; gyroscope 'none'; magnetometer 'none'; microphone 'none'; payment 'none'; usb 'none'"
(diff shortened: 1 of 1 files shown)


vikaspotluri123:
- includeSubdomains wasn't included because the only subdomain in use was www which redirects to the naked domain. I don't think it makes a difference for us either way - if we do add it, it adds a few bytes to the header, that's it

- The time was set lower just in case something didn't work out; increasing it is fine.

aslafy-z:
The two edits are required to integrate the hstspreload list which is loaded by modern browsers. That's why I added `includeSubDomains` even if there is no.

swapagarwal:
Thanks @aslafy-z! 👍

swapagarwal:
Submitted to the list; pending review: https://hstspreload.org/?domain=devswag.io

PR #212, opened 2018-12-08
Fix watch task

aslafy-z:

Follow up #170

Cachebust is still executed in the `build` task pipeline. We need to explicitly skip in while in dev.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -114,9 +114,14 @@ gulp.task('clean:styl', () => del('dist/assets/css/*'));
 gulp.task('clean:js', () => del('dist/assets/js/*'));
 gulp.task('clean:assets', gulp.parallel('clean:styl', 'clean:js'));
 gulp.task('clean:pug', () => del('dist/index.html'));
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -160,8 +164,8 @@ gulp.task('webserver', () => {
 
 gulp.task('watch', () => {
 	gulp.watch('src/pug/**/*.pug', gulp.series('pug'));
-	gulp.watch('src/styl/**/*.styl', gulp.series('clean:styl', 'styl'));
-	gulp.watch('src/js/*.js', gulp.series('clean:js', 'js'));
+	gulp.watch('src/styl/**/*.styl', gulp.series('styl'));
```
aslafy-z: Disabling these clean tasks is not required.
Give me your point about it!

aslafy-z: @vikaspotluri123 wdyt?

vikaspotluri123: It's not because watch should only run in dev mode -> cachebusting is disabled

PR #218, opened 2018-12-10
Hotfix: set production mode for npm 'build' script

aslafy-z:
Fix netlify build to use production workflow


aslafy-z:
This has been lost while merging. cachebusted assets are back!

PR #220, opened 2018-12-11
Fix/js deps to body

aslafy-z:

From [https://devswag.io](https://devswag.io): Line 134

```
</body><script src="https://unpkg.com/mobius1-selectr@2.4.4/dist/selectr.min.js"></script><script defer='defer' src='https://d33wubrfki0l68.cloudfront.net/js/b81343513a8edc4d23853dd8f7fcb6d0f6114c8c/assets/js/index-087cc6a5d5.js'></script></html>
```

These scripts have to be within the body section. Moving them back in it.

### DIFF
--- a/src/pug/index.pug
+++ b/src/pug/index.pug
@@ -10,6 +10,6 @@ html(lang='en')
 
 		include ./includes/body.pug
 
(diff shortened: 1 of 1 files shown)


vikaspotluri123:
> These scripts have to be within the body section.

What's the reasoning for this? I wasn't aware of any such rule 😬

PR #222, opened 2018-12-11
Use jsDeliver instead of unpkg as CDN provider

aslafy-z:

Unpkg CDN is slower than jsDeliver, see https://www.cdnperf.com/

They also have a uge [sponsors list](https://www.jsdelivr.com/sponsors) compared to [unpkg one](https://unpkg.com/#/about)


plibither8:
LGTM! Merging :+1::smile:

PR #223, opened 2018-12-11
Implement Image lazyloading

aslafy-z:

Native lazy loading will come in future versions of web browsers: https://css-tricks.com/a-native-lazy-load-for-the-web-platform/

Since then, I included the https://github.com/ApoorvSaxena/lozad.js module, the most tiniest and up to date module I found. It has no dependencies, <1k, and pretty fast.

I used this image as a placeholder, feel free to propose another (it could be the logo when we have one): https://loading.io/spinner/heart/-beating-heart-preloader

spinner

Once it's merged, we can add [devSwag](https://devswag.io) to lozad users list:
https://github.com/ApoorvSaxena/lozad.js/#more-than-3000-sites-are-using-lozadjs-heres-few-of-them


plibither8:
Instead of showing nothing to those users that have disabled JS, let's disable lazyloading? At least the users will be able to see the content.

aslafy-z:
@plibither8 Fixed! I forgot to add the logic to support that case :sweat_smile:

aslafy-z:
[`lozad`](https://github.com/ApoorvSaxena/lozad.js/) **does** load images on-the-fly to decrease wasted bandwidth using intersection observer. Gonna do some more benches tomorrow, but I think it worth the load of the library.

vikaspotluri123:
It might be running it too early because I only see beating hearts once

I need more beating hearts! 😝

aslafy-z:
I guess it's too early for `loading` attr :(
https://caniuse.com/#feat=loading-lazy-attr

aslafy-z:
#587 does this.

PR #224, opened 2018-12-11
Project wide editorconfig lint

aslafy-z:

Use https://github.com/jedmao/eclint to check and fix lint respecting rules from `.editorconfig` file.

Added to the lint-staged pipeline.

`LICENSE` is set to be ignored as it does not respect any rules, we can also add a specific rule for it instead of ignoring it.


swapagarwal:
Why moving away from XO?

aslafy-z:
We are not! `xo` is a `js` only linter. 
We use eclint to lint all other files (pug, styl, html, json, ...)

swapagarwal:
Oops. My bad! 😅

.editorconfig
```diff
@@ -6,3 +6,16 @@ end_of_line = lf
...
+block_comment_start = /*
+block_comment = *
+block_comment_end = */
+
+[*.md]
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = false
+
+[{*.json,*.yml}]
```
vikaspotluri123: and toml!

src/styl/imports/base.styl
```diff
@@ -4,7 +4,7 @@ html, body, :root
 	width 100%
 	height 100%
 	background #fff
-	font-family 'Cabin', -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif	
+	font-family 'Cabin', -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif
```
vikaspotluri123: This might cause a merge conflict for #225 - if you can, skip this here and fix it there 😊

package.json
```diff
@@ -63,7 +64,12 @@
   },
   "lint-staged": {
```
vikaspotluri123: General question: is this absolutely required? I know there are lots of repos that do this, but (for me personally) it wastes a lot of time because I have to wait for lint to run until I can commit + push to remotes and it makes me not want to commit because there's too much friction

plibither8: I agree. I would prefer a pre-push hook rather than pre-commit

vikaspotluri123: With CI, failing lint prevents merge

plibither8: But CI is after pushing changes, right? And CI evaluates the last commit.

aslafy-z: Something like this: https://github.com/theenadayalank/lint-prepush?

plibither8: Why can't we just do this:
```
"pre-push": "npm run lint:fix"
```

aslafy-z: If the task returns without any error (even if it linted some files, unstaged so), it will continue to push without creating a fix commit.

vikaspotluri123: > The `lint-staged` task runs really fast on my computer, is it slow for you?

Yep, adds 3-5 seconds, even though I have an ssd 😭

aslafy-z:
Can I merge this?

vikaspotluri123:
👍

package.json
itaisteinherz: I usually see projects checking for linting errors only on CI. If you decide on going in this route, I think it's preferable to have linting done automatically on push instead of on every commit, since the latter makes the entire iteration process slower and more annoying.
Regarding creating a `lint-fix` commit automatically on push / commit, that doesn't seem like a good idea to me, since many times I've had auto-fixable linting errors which I needed to fix manually as that was more appropriate.

These are just my opinions and observations from working on other projects, though.

PR #225, opened 2018-12-11
Non blocking fonts loading

aslafy-z:

- Selfhost Cabin font downloaded from https://google-webfonts-helper.herokuapp.com/fonts/cabin?subsets=latin
- Use [`font-display`](https://caniuse.com/#search=font-display) to autoswitch to the font when loaded.
- Tweak fallback fonts (CSStricks)


gulpfile.js
```diff
@@ -67,6 +67,11 @@ gulp.task('styl', () => {
 		.pipe(gulp.dest('dist/assets/css'));
 });
 
+gulp.task('fonts', () => {
```
vikaspotluri123: Could we merge this into images and create a `binaries` task? both the fonts and img folder won't change  frequently and adding a task for this seems a bit extra

aslafy-z: `img` tasks has some more processing in it.. I can rename the task but I'm not sure if we should pass all the binaries through `gulp-responsive`.

vikaspotluri123: Since the images are static, I think we should optimize it once and just use it, rather than pointlessly reoptimizing it. We can always use blame / history to get previous versions of the one image in there 😉

aslafy-z:
Thanks guys, lets merge this one! :tada:

Issue #103, opened 2018-10-09
Images: force jpeg?


vikaspotluri123:
Refs https://github.com/swapagarwal/swag-for-dev/pull/93#discussion_r223505178

Currently images are forced to be jpegs when they're processed. Given the source hopefully put thought into image type, is it worth forcing the image type to be jpeg?

Even though the site is tiny (500k), what about modern formats, such as progressive jpegs or webp?

aslafy-z:
Just to clarify, images are not forced to be jpegs but they we force the conversion to jpegs (progressive jpegs, see [#93 (diff)](https://github.com/swapagarwal/swag-for-dev/pull/93/files#diff-6eb60389fb540a5c670ae4d43af865a9R19)).

jpeg2000: only safari: https://caniuse.com/jpeg2000
webp: only chrome/opera: https://caniuse.com/webp

webp is coming to firefox and ie but it's very fresh:
https://www.cnet.com/news/firefox-to-support-googles-webp-image-format-for-a-faster-web/
https://www.cnet.com/news/googles-webp-image-format-finally-finds-a-browser-ally-microsoft-edge/

still nothing for safari...

aslafy-z:
WebP support has been integrated. Closing this from now.

PR #226, opened 2018-12-11
Generate WebP images

aslafy-z:

- Generate some webp images using `gulp-responsive`.
- Use `<picture>` and `srcset` to reference it optionally (see https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images)

Ref #103

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -101,7 +101,7 @@ gulp.task('swag-img:download', () => {
 
 gulp.task('swag-img:optimize', () => {
 	const imageSettings = {
-		name: '*.jpg',
+		name: '*.jpeg',
 		height: 300,
 		flatten: true
 	};
@@ -111,10 +111,18 @@ gulp.task('swag-img:optimize', () => {
 			imageSettings,
 			// WebP output
 			Object.assign({
-				rename: i => `${i.basename}.webp`
+				rename: i => `${i.origin}.webp`
 			}, imageSettings)
 		], RESIZE_OPTS))
-		.pipe(gulp.dest('dist/assets/swag-img'));
+		.pipe(gulp.dest('dist/assets/swag-img'))
+		.on('end', () => {
+			swagList.forEach(swag => {
+				swag.image = {
+					jpeg: swag.image,
+					webp: `${swag.image}.webp`
+				};
+			});
+		});
 });
 
 gulp.task('swag-img', gulp.series('swag-img:clean', 'swag-img:download', 'swag-img:optimize'));
@@ -130,7 +138,6 @@ gulp.task('cachebust', cb => {
 	if (!PRODUCTION) {
 		return cb();
 	}
-
 	const basePath = 'dist/assets';
 	const bustedFiles = [
 		'dist/assets/css/*',
@@ -150,14 +157,15 @@ gulp.task('cachebust', cb => {
 		.on('end', () => {
 			delete require.cache[require.resolve(REV_PATH)];
 			manifest = require(REV_PATH);
-
 			swagList.forEach(swag => {
-				const filename = `swag-img/${swag.image.split('/').pop()}`;
-				if (!manifest[filename]) {
-					console.warn('Unable to find image in manifest:', filename);
-					return;
-				}
-				swag.image = `/assets/${manifest[filename]}`;
+				Object.entries(swag.image).forEach(([extension, filename]) => {
+					filename = `swag-img/${filename.split('/').pop()}`;
+					if (!manifest[filename]) {
+						console.warn('Unable to find image in manifest:', filename);
+						return;
+					}
+					swag.image[extension] = `/assets/${manifest[filename]}`;
+				});
 			});
 
 			cb();
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -100,13 +100,20 @@ gulp.task('swag-img:download', () => {
 });
 
 gulp.task('swag-img:optimize', () => {
+	const imageSettings = {
+		name: '*.jpg',
+		height: 300,
+		flatten: true
+	};
 	return gulp.src('dist/assets/swag-img/*')
-		.pipe(responsive([{
-			name: '*',
-			height: 300,
-			format: 'jpeg',
-			flatten: true
-		}], RESIZE_OPTS))
+		.pipe(responsive([
+			// JPEG output
+			imageSettings,
+			// WebP output
+			Object.assign({
+				rename: i => `${i.basename}.webp`
+			}, imageSettings)
+		], RESIZE_OPTS))
 		.pipe(gulp.dest('dist/assets/swag-img'));
 });
 
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -110,10 +110,7 @@ gulp.task('swag-img:optimize', () => {
 	}, jpegOutputConfig);
 
 	return gulp.src('dist/assets/swag-img/*')
-		.pipe(responsive([
-			jpegOutputConfig,
-			webpOutputConfig
-		], RESIZE_OPTS))
+		.pipe(responsive([jpegOutputConfig, webpOutputConfig], RESIZE_OPTS))
 		.pipe(gulp.dest('dist/assets/swag-img'))
 		.on('end', () => {
 			swagList.forEach(swag => {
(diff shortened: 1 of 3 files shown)


vikaspotluri123:
So I haven't looked into webp support a bunch, but from what I remember reading a while ago, there's some server response involved with sending webp - I'm guessing Netlify can handle this because SmashingMag uses webp, but I'm curious if you have any insights 😇

aslafy-z:
I thought `webp` whould work out of the box for most web servers but some might not set the `Content-Type` correctly because it's mime type is *still* not listed in the official [IANA mime list](https://www.iana.org/assignments/media-types/media-types.xhtml). Still, they all uses `image/webp` as the mime type for webp.

See https://bugzilla.mozilla.org/show_bug.cgi?id=600919#c139 and below.

Netlify seems to accept it:

```
$ curl -I https://deploy-preview-226--peaceful-chandrasekhar-efde8e.netlify.com/assets/swag-img/devrant-26d3018fd4.webp 
HTTP/2 200 
accept-ranges: bytes
content-length: 8678
content-type: image/webp           <===== YEAH
date: Wed, 12 Dec 2018 20:45:44 GMT
etag: "bc4160e1173cdf61f9e516bbd1a7638a-ssl"
feature-policy: accelerometer 'none'; camera 'none'; geolocation 'none'; gyroscope 'none'; magnetometer 'none'; microphone 'none'; payment 'none'; usb 'none'
referrer-policy: strict-origin-when-cross-origin
strict-transport-security: max-age=31536000; includeSubDomains; preload
x-content-type-options: nosniff
x-frame-options: SAMEORIGIN
x-nf-srv-version: 72b8ce8c0aacc79c990e81dd93dc325fb603aadd
x-robots-tag: noindex
x-xss-protection: 1; mode=block
age: 2
server: Netlify
cache-control: public max-age=63072000
x-nf-request-id: 2e347582-98d5-4571-ae26-84f5649895f9-4004685
```

vikaspotluri123:
I've gotten some sleep, and I remember the issue - browsers that don't support webp!

aslafy-z:
See https://github.com/swapagarwal/swag-for-dev/issues/103#issuecomment-428248731 

Just seen this article, from today! https://www.androidpolice.com/2018/12/12/firefox-beta-65-adds-support-for-webp-images/

vikaspotluri123:
We have too many Safari users to ignore them 😕

aslafy-z:
Have a look to https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images. The goal is not to drop jpeg but add webp support for browsers that supports it.

vikaspotluri123:
Gotcha, didn't realize there was a plan to switch from `img` to `picture` (or something along those lines) 👍

aslafy-z:
Still have to cleanup the webp filename, but it works!

gulpfile.js
```diff
@@ -147,14 +154,15 @@ gulp.task('cachebust', cb => {
...
 			manifest = require(REV_PATH);
-
 			swagList.forEach(swag => {
-				const filename = `swag-img/${swag.image.split('/').pop()}`;
-				if (!manifest[filename]) {
-					console.warn('Unable to find image in manifest:', filename);
-					return;
-				}
-				swag.image = `/assets/${manifest[filename]}`;
+				Object.entries(swag.image).forEach(([extension, filename]) => {
```
vikaspotluri123: Wouldn't you be trying to destructure a string here?

gulpfile.js
```diff
@@ -98,14 +98,29 @@ gulp.task('swag-img:download', () => {
 });
 
 gulp.task('swag-img:optimize', () => {
+	const imageSettings = {
```
vikaspotluri123: let's rename this to `jpegOutputConfig` for clarity

gulpfile.js
```diff
@@ -98,14 +98,29 @@ gulp.task('swag-img:download', () => {
...
-			height: 300,
-			format: 'jpeg',
-			flatten: true
-		}], RESIZE_OPTS))
-		.pipe(gulp.dest('dist/assets/swag-img'));
+		.pipe(responsive([
+			// JPEG output
+			imageSettings,
+			// WebP output
+			Object.assign({
```
vikaspotluri123: Might be worth creating a variable for this outside of here for clarity

src/pug/includes/body.pug
```diff
@@ -53,7 +53,11 @@ include ./fork.pug
                         each swagTag in swag.tags
                             span!=swagTag
                     div.flex.img-container
-                        img(src=swag.image alt=swag.name + " swag you can get")
+                        picture
```
vikaspotluri123: 💃

gulpfile.js
aslafy-z: Nope, see here: https://github.com/swapagarwal/swag-for-dev/pull/226/commits/1ae989bf18b8e03b9fa3fba59c28051cfa93dd29#diff-b9e12334e9eafd8341a6107dd98510c9R118. Just switched to `images` to be clearer.

src/pug/includes/body.pug
aslafy-z: Did I miss something?

vikaspotluri123: Nope, it's a good thing!

gulpfile.js
vikaspotluri123: whoops, I thought you were using `Object.values` - I think that would be better here?

aslafy-z: I reuse the extension from the object keys. If we want to use Object values I'll have to compute it from filename like so: 

```diff
diff --git a/gulpfile.js b/gulpfile.js
index af6b475..668cf4c 100644
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -155,8 +155,9 @@ gulp.task('cachebust', cb => {
                        delete require.cache[require.resolve(REV_PATH)];
                        manifest = require(REV_PATH);
                        swagList.forEach(swag => {
-                               Object.entries(swag.images).forEach(([extension, filename]) => {
+                               Object.values(swag.images).forEach(filename => {
                                        filename = `swag-img/${filename.split('/').pop()}`;
+                                       const extension = filename.split('.').pop();
                                        if (!manifest[filename]) {
                                                console.warn('Unable to find image in manifest:', filename);
                                                return;
```

aslafy-z: Gonna keep `Object.entries` as it's fits best

gulpfile.js
```diff
@@ -98,14 +98,30 @@ gulp.task('swag-img:download', () => {
...
+
 	return gulp.src('dist/assets/swag-img/*')
-		.pipe(responsive([{
-			name: '*',
-			height: 300,
-			format: 'jpeg',
-			flatten: true
-		}], RESIZE_OPTS))
-		.pipe(gulp.dest('dist/assets/swag-img'));
+		.pipe(responsive([
```
vikaspotluri123: This could probably fit in one line

gulpfile.js
```diff
@@ -140,14 +155,15 @@ gulp.task('cachebust', cb => {
...
-					return;
-				}
-				swag.image = `/assets/${manifest[filename]}`;
+				Object.entries(swag.images).forEach(([extension, fileName]) => {
+					fileName = `swag-img/${fileName.split('/').pop()}`;
+					if (!manifest[fileName]) {
+						console.warn('Unable to find image in manifest:', fileName);
+						return;
+					}
+					swag.images[extension] = `/assets/${manifest[fileName]}`;
```
vikaspotluri123: didn't notice this earlier, gotcha!

gulpfile.js
```diff
@@ -100,14 +100,27 @@ gulp.task('swag-img:download', () => {
...
-		.pipe(responsive([{
-			name: '*',
-			height: 300,
-			format: 'jpeg',
-			flatten: true
-		}], RESIZE_OPTS))
-		.pipe(gulp.dest('dist/assets/swag-img'));
+		.pipe(responsive([jpegOutputConfig, webpOutputConfig], RESIZE_OPTS))
+		.pipe(gulp.dest('dist/assets/swag-img'))
+		.on('end', () => {
```
vikaspotluri123: I'm pretty sure you need to use a callback if you add an `on('end')` listener because you run into race conditions

plibither8:
Looks good :+1:, I think we can merge :)

vikaspotluri123:
Thanks @aslafy-z! I was served webp in my browser ⚡ 🎉

Issue #79, opened 2018-10-03
Add automated tests


aslafy-z:
Need to check that

- code is linted
- the site builds
- data.json is well formated 
- data.json links are not dead 

What else ?

swapagarwal:
I was thinking along those lines too!

aslafy-z:
I seen this,

![2018-10-04-124736_174x47_scrot](https://user-images.githubusercontent.com/8191198/46469440-b4ca8c00-c7d3-11e8-942f-f6638d98bc22.png)

Are you using it Codeship ? Should we use it or another one (TravisCI, CircleCI, another one) ?
Could you grant me a read access to it (or turn it public) ?

aslafy-z:
Some visual testing materials:
- https://percy.io/
- https://github.com/apps/reg-suit
- https://applitools.com/
- https://github.com/mojoaxel/awesome-regression-testing

swapagarwal:
Yes, I was using CodeShip to test that the links are not dead. Since now we have a visual testing requirement, we'll need to explore the various options available. 😃

swapagarwal:
Copied from #82 
`We need to have a test framework (not needed but better), add a test case for each platform we want to test on and use a CI to trigger them. As seen there: https://www.browserstack.com/automate/node#testing-frameworks`

swapagarwal:
@vikaspotluri123 See if you get time?

swapagarwal:
Closed in #228

vikaspotluri123:
Did we want to add visual testing?

swapagarwal:
It's probably over-engineering at this point. 😅
The most active time is October only, and the rest of the year is pretty much unhappening (as we saw last year!).

PR #115, opened 2018-10-11
WIP: Add CI

aslafy-z:
Implement automatic build for local & docker installations using Circle CI. 

This is based on https://github.com/aslafy-z/swag-for-dev/tree/add-dockerfile/
I'll rebase it on master when #76 will be merged 

**TODO**
- [ ] Fix docker caching
- [ ] PR support
- [ ] Use browserstack ?
- [ ] Push artefacts (screenshots) to the PR
- [ ] Deploy to netlify when it succeed
- [ ] You tell me

PR #228, opened 2018-12-12
👷‍♂️ add functional tests

vikaspotluri123:
- Refs #79

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ node_js:
 sudo: false
 env:
   - TEST_SUITE=lint
-  - TEST_SUITE=mocha
+  - TEST_SUITE=data-schema
 
 cache:
   directories:
@@ -15,8 +15,6 @@ cache:
 install:
   - npm ci
 
-script: >
-    if [ "$TEST_SUITE" == "lint" ];
-        then npm run lint;
-        else npm run test;
-    fi
+script:
+    - if [ "$TEST_SUITE" == "lint" ]; then npm run lint; fi
+    - if [ "$TEST_SUITE" == "data-schema" ]; then npm run test; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,19 @@
+# Formatting set by travis CLI
 language: node_js
-
 node_js:
-  - "10"
-
+- '10'
 sudo: false
 env:
-  - TEST_SUITE=lint
-  - TEST_SUITE=data-schema
-
+- TEST_SUITE=lint
+- TEST_SUITE=data-schema
 cache:
   directories:
-    - node_modules
-
+  - node_modules
 install:
-  - npm ci
-
+- npm ci
 script:
-    - if [ "$TEST_SUITE" == "lint" ]; then npm run lint; fi
-    - if [ "$TEST_SUITE" == "data-schema" ]; then npm run test; fi
+- if [ "$TEST_SUITE" == "lint" ]; then npm run lint; fi
+- if [ "$TEST_SUITE" == "data-schema" ]; then npm run test; fi
+notifications:
+  slack:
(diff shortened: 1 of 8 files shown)


vikaspotluri123:
Mocha is tried and true, but the main reason I use it is because I know how to easily and quickly bootstrap it 😁

itaisteinherz:
From my experience testing Node.js projects, it's usually best to run linting and test with Mocha in a single build. Linting is usually a minor test meant to point out basic issues in the code, and there shouldn't be a separate testing environment just for it.

What I suggest is that the `test` script (defined in `package.json`) should be `npm run lint && mocha test/functional/data.spec.js`, and then we can remove the additional `env` and `script` configurations from `.travis.yml`.

vikaspotluri123:
The reason for parallelizing builds is speed - running lint can take 5-10 seconds, which adds to the overall time for running the tests, especially the integrity tests since every URL has to be tested. Given we have the resources, I chose to improve CI run time in this method.

I don't entirely follow as to why running parallel tests is bad / not good - could you please expand on this?

itaisteinherz:
First of all, I don't really consider linting a test, but a general check before running unit tests and such. In this case, we're testing all of the URLs in the same build, and so running XO before the integration tests doesn't add that much of an overhead.
Running more than one build in this case would be excessive and redundant, in my opinion.

vikaspotluri123:
@itaisteinherz I'm going to leave it the way it is for now. The main reason is the test suite (currently) is doing 2 separate things

1. Validating the `data.json` file 
   - check that every object follows the required schema
   - check that all the URLs are valid
2. Make sure the code style matches our required guidelines

While 1 might be impacted by 2, the impact should be minimal

.travis.yml
```diff
@@ -0,0 +1,19 @@
...
+
+cache:
+  directories:
+    - node_modules
+
+install:
+  - npm ci
+
+script:
+  - if [ "$TEST_SUITE" == "lint" ]; then npm run lint; else npm run test; fi
```
aslafy-z: I think we can use one line per test suite and no else so it's easier to read

test/functional/data.spec.js
```diff
@@ -0,0 +1,52 @@
+const got = require('got');
+const {expect} = require('chai');
+
+let data;
+
+function checkURL(url, head = false) {
+	const method = head ? 'head' : 'get';
+	return got[method](url, {throwHttpErrors: false}).then(({statusCode}) => {
+		if (statusCode === 403) {
```
aslafy-z:
```suggestion
		if (!head && statusCode === 403) {
```

I think we should add a guard, otherwise it could infinite loop and crash

.travis.yml
vikaspotluri123: Updated - is this what you were thinking?

aslafy-z: I was thinking something like this instead:

```bash
if [ "$TEST_SUITE" = "lint" ]; then npm run lint; fi
if [ "$TEST_SUITE" = "mocha" ]; then npm run test; fi
```

Yash-Garg:
Can add build notifications too on failure in my opinion!

```shell
notifications:
  email:
    recipients:
      - EMAIL
    on_success: never
    on_failure: always
```

aslafy-z:
I'd go for slack notifications as we already have some. I think it should
trigger an error anyway if the Travis GitHub Check fails.


Le sam. 26 janv. 2019 à 20:28, Vikas Potluri <notifications@github.com> a
écrit :

vikaspotluri123:
Worked on the first try 😎

Yash-Garg:
Great 👌

test/functional/data.spec.js
```diff
@@ -0,0 +1,52 @@
...
+			expect(datum.reference).to.be.a('string');
+			expect(datum.image).to.be.a('string');
+			expect(datum.dateAdded).to.match(/20\d\d-[0-1][\d]-[0-3][\d]T\d\d:\d\d:\d\d\.000Z/);
+			expect(datum.tags).to.be.an('Array').with.length.greaterThan(0);
+		});
+	});
+
+	describe('valid images and references', function () {
+		data = require('../../data.json');
+		delete require.cache[require.resolve('../../data.json')];
```
aslafy-z: This should be done before calling require. It might be good to add it in the `swag-for-dev.data.json is valid` also

vikaspotluri123: I don't think this delete is super necessary now that I think about it. I had initially to make the 2 blocks somewhat idempotent, but if the require fails here, it would fail there as well so it's kind of useless - what do you think?

aslafy-z: yep in fact it's not necessary from now. let's drop the second require & cachebust

vikaspotluri123: Done 👍

test/functional/data.spec.js
```diff
@@ -0,0 +1,49 @@
...
+		if (!head && statusCode === 403) {
+			return checkURL(url, true);
+		}
+
+		expect(statusCode).to.equal(200);
+	});
+}
+
+describe('swag-for-dev', function () {
+	it('data.json is valid', function () {
```
aslafy-z: I thing you should keep the require in the test loop, either by adding it to `data.json is valid` or creating a first `it` case that loads it to data? So if we use reporters in the future it won't crash and send invalid output?

vikaspotluri123: If the tests fail, data shouldn't be sent to external reporters, right? Especially in PRs 🤔

aslafy-z: My bad, your're right

aslafy-z:
I have a last concert about Travis CI, it's not so extendable as other to run dockers and services like Circle CI or Codeship.

I did a first try with Circle CI in #115 
https://github.com/swapagarwal/swag-for-dev/pull/115/files#diff-1d37e48f9ceff6d8030570cd36286a61
Maybe we can switch the whole to Circle CI so it can be used?

vikaspotluri123:
I have no issue with switching CI providers, but I only think we should switch if it's affecting the flow - what specifically does CircleCI that Travis doesn't? It's possible to run docker tests in Travis by installing docker, so I don't fully understand what the issue is 🙂

aslafy-z:
We'll see if we stick to Travis later but it will do from now! Go ahead :wink:

aslafy-z:
Hi @vikaspotluri123! Any news on that one?

vikaspotluri123:
I think we're just waiting for an approval... I'll rebase once it's approved 😃

swapagarwal:
`LGTM!`
@vikaspotluri123 Can you rebase and merge it? 😇

vikaspotluri123:
Removing Mullvad since their page is 404'ing

vikaspotluri123:
It seems like Husky isn't running the proper eclint / XO command, since lint passes on prepush but fails in Travis 🤔

PR #231, opened 2018-12-14
Make Selectr look like other select inputs

aslafy-z:

- Hide Selectr until it's loaded
- Style Selectr
- Disable search feature
- Use Selectr arrow impl for other select inputs

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -14,6 +14,7 @@ const sort = {
 const contentEl = document.getElementById('content');
 const filterInput = document.getElementById('filter');
 const sortingInput = document.getElementById('sorting');
+const tagsSelect = document.getElementById('tags');
 
 const activateElements = els => Array.from(els).forEach(node => node.classList.add(ACTIVE_CLASS));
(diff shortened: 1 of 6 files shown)


src/js/index.js
```diff
@@ -87,9 +89,12 @@ function cascade(force = false) {
 window.addEventListener('load', () => {
 	selectr = new Selectr('#tags', {
 		multiple: true,
+		searchable: false,
 		placeholder: 'Choose tags...',
 		data: window.swagTags.map(tag => ({value: tag, text: tag}))
 	});
+	tagsPlaceholderSelect.style.display = 'none';
```
vikaspotluri123: I think we should hide the tags search and enable it when Selectr loads - thoughts?

vikaspotluri123: Inlining the tag selector is done in #159

vikaspotluri123:
I think the one minor consistency issue is the options shouldn't be cursor: `pointer` , rather it should be `default`

PR #232, opened 2018-12-16
added Alligator swag to data.json

olessiap:

hey there! I added the alligator info to data.json. I also signed up to write for the platform to get the swag. I'll update here once that happens!

### DIFF
--- a/data.json
+++ b/data.json
@@ -9,6 +9,15 @@
     "tags": ["clothing", "device"]
   },
   {
+    "name": "Alligator.io",
+    "difficulty": "hard",
+    "description": "Write a blog post on frontend development topics, get some cool Alligator stickers",
+    "reference": "https://alligator.io/write-for-us/",
+    "image": "https://pbs.twimg.com/profile_images/622200384455442432/hf71Z251_400x400.png",
+    "dateAdded": "2018-12-15T06:03:00.000Z",
+    "tags": ["stickers"]
+  },
+  {
     "name": "Aras",
     "difficulty": "medium",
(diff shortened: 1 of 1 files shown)


data.json
```diff
@@ -8,6 +8,15 @@
     "dateAdded": "2018-02-18T06:03:00.000Z",
     "tags": ["clothing", "device"]
   },
+  {
+    "name": "Alligator.io",
+    "difficulty": "hard",
+    "description": "Write a blog post on frontend development topics, get some cool Alligator stickers",
+    "reference": "https://alligator.io/write-for-us/",
+    "image": "https://pbs.twimg.com/profile_images/622200384455442432/hf71Z251_400x400.png",
+    "dateAdded": "2018-12-15T06:03:00.000Z",
```
vikaspotluri123: This should be `2018-12-09T07:08:00.000Z` since we (devSwag) were notified then

olessiap: @vikaspotluri123, hello! I fixed the timestamp. Can you please tell me where you see it? I see that the issue was opened 7 days ago, is that what you're referring to? 
:)

vikaspotluri123: Yep, that's exactly right!

data.json
```diff
@@ -8,6 +8,15 @@
     "dateAdded": "2018-02-18T06:03:00.000Z",
     "tags": ["clothing", "device"]
   },
+  {
+    "name": "Alligator.io",
+    "difficulty": "hard",
+    "description": "Write a blog post on frontend development topics, get some cool Alligator stickers",
```
aslafy-z: Let's add some ending ponctuation to keep consistency with other swag items

olessiap: @aslafy-z whoops, wow good eye. I fixed it!

vikaspotluri123:
Thanks again @olessiap!

PR #247, opened 2019-01-12
Set default text color to black

aslafy-z:

Here's how the select renders in Firefox:

Applied a patch to set default text color to black (#000).

### DIFF
--- a/src/styl/imports/base.styl
+++ b/src/styl/imports/base.styl
@@ -22,7 +22,6 @@ html, body, :root
 	-webkit-font-smoothing antialiased
 	-moz-osx-font-smoothing grayscale
 	-webkit-overflow-scrolling touch
-	color #000
 
 code
 	padding 1px 5px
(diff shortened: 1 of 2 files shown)


vikaspotluri123:
Looks like ubscri.be is creating issues here as well /cc @itaisteinherz

aslafy-z:
`select` elements doesn't look to inherit values from selector `html, body, :root`.
Moving color rule to `select` selector.

aslafy-z:
Yeah, seen that. Just find out I'm able to load the page in a private browser session (or another browser). Seems like it's linked to some cookies or something. Even with blocking every external resource with uBlock Origin, the page still hangs after showing up.

aslafy-z:
Fixed for Firefox.

vikaspotluri123:
I just blocked upscribe at the system level, and the page still hangs. I don't want to merge without determining the root cause of the problem. Haven't had a chance to compare what the color difference is like because of the hang

aslafy-z:
@vikaspotluri123 give an eye to #159

aslafy-z:
Oh, so it might be linked to my custom GTK theme.

aslafy-z:
Not sure if it's legit thing to change the common color for selects. It might be a problem with my GTK theme instead.

Right, downsides of theming native inputs. If I disable the theming it shows like this: 
2019-01-14-125725_455x54_scrot
Definitively linked to my theme. Closing this as doesn't look to be linked with devSwag.

PR #248, opened 2019-01-12
Fix typos

aslafy-z:

Moved typo fixes out of #159 to this PR.
Thanks to @itaisteinherz!


itaisteinherz:
Sure! 🙂 I added these to #159 since they were small (and minor) changes which I didn't think required another PR, but this also works.

PR #250, opened 2019-01-14
feat: lazy load CSS

Larsklopstra:

Added a function in JavaScript to lazyload the stylesheets and included the critcal path css in the header. There's also a fallback if the user has noscript enabled.


src/js/index.js
```diff
@@ -5,20 +5,23 @@
  */
 const ACTIVE_CLASS = 'visible';
 const sort = {
-	ALPHABETICAL_ASCENDING: (a, b) => a.dataset.name > b.dataset.name ? 1 : -1,
-	ALPHABETICAL_DESCENDING: (a, b) => a.dataset.name < b.dataset.name ? 1 : -1,
-	DIFFICULTY_ASCENDING: (a, b) => a.dataset.difficulty > b.dataset.difficulty ? 1 : -1,
-	DIFFICULTY_DESCENDING: (a, b) => a.dataset.difficulty < b.dataset.difficulty ? 1 : -1
+	ALPHABETICAL_ASCENDING: (a, b) => (a.dataset.name > b.dataset.name ? 1 : -1),
```
aslafy-z: Please lint your JS code with `npm run lint:fix`

src/pug/includes/head.pug
```diff
@@ -10,9 +10,5 @@ meta(name="description" content="swag opportunities for developers")
...
...;font-size:.9rem}section.main .content{max-width:1100px;flex-wrap:wrap;align-items:flex-start;justify-content:space-around;margin:0 -10px;width:100%}section.main .content .item{width:100%;max-width:350px;padding:30px 10px;border-top:1px solid #ddd;display:none}@media only screen and (max-width:500px){section.main .content .item{max-width:none}}section.main .content .item .title{justify-content:space-between}section.main .content .item .title h1{font-size:2rem}section.main .content .item .title .difficulty{height:30px;width:30px;border-radius:50%;border:2px solid #000}section.main .content .item .title .difficulty.hard{background:#ff0009}section.main .content .item p{font-size:1.2rem;line-height:1.4;color:#222}section.main .content .item p.swag{margin:5px -3px 10px -3px}section.main .content .item p.swag span{background:#737373;color:#fff;border-radius:3px;margin:0 3px;padding:2px 5px;font-size:1rem}section.main .content .item img{height:100%;width:auto}section.main .content .item .img-container{height:300px;width:100%;overflow:hidden;padding-bottom:15px}.sr-only{position:absolute!important;clip:rect(1px 1px 1px 1px);clip:rect(1px,1px,1px,1px)}#content .item.visible{display:initial}
```
aslafy-z: This CSS is being generated by gulp so the way to do that would be passing it to the pug template.
I think we could keep our workflow for our CSS but maybe  lazyload the selectr CSS.

src/js/index.js
```diff
@@ -85,7 +104,12 @@ function cascade(force = false) {
 	}
 }
 
-window.addEventListener('load', () => {
+window.addEventListener('DOMContentLoaded', () => {
+	lazyLoadStyleSheet('/assets/css/index.css');
```
aslafy-z: This is the path to the busted assets only in dev mode. It might be changed one layer upper by netlify also so we shouldn't reference it directly like this.

src/pug/includes/head.pug
```diff
@@ -10,9 +10,5 @@ meta(name="description" content="swag opportunities for developers")
 link(rel="icon" type="image/png" href="/assets/img/logo.png")
 link(rel="apple-touch-icon" size="128x128" href="/assets/img/logo.png")
 
-link(rel="stylesheet" href=bustedAssets.css)
-link(rel='stylesheet' href='https://cdn.jsdelivr.net/npm/mobius1-selectr@2.4.8/dist/selectr.min.css')
-
-noscript
-	style.
-		.buttons {display: none}
```
aslafy-z: We still need this!

src/js/index.js
```diff
@@ -85,7 +104,12 @@ function cascade(force = false) {
 	}
 }
 
-window.addEventListener('load', () => {
+window.addEventListener('DOMContentLoaded', () => {
```
aslafy-z: +1 on using `DOMContentLoaded` instead of `load` 

Reference: https://stackoverflow.com/questions/2414750/difference-between-domcontentloaded-and-load-events

src/pug/includes/head.pug
Larsklopstra: This is the critical path CSS generated by a third party site

src/js/index.js
Larsklopstra: Will the path always be `assets/...`? Maybe using `location.hostname` will solve it, otherwise I can make a fake `link` with a `data-href` attribute which will be swapped when the page is ready :)

src/pug/includes/head.pug
aslafy-z: We might be able to extract our CSS critical path with https://github.com/addyosmani/critical within our gulp pipeline.

src/js/index.js
aslafy-z: I'd go with `data-href` on the `link` so we keep our dynamic filenames in the html (`bustedAssets.css` variable).

src/pug/includes/head.pug
vikaspotluri123: Critical CSS is normally a great idea, but given we're using H2 Multiplexing and the `index-{hash}.css` file is 2k (ish), I don't think it's really necessary. This would be a premature optimization. Looking at the lighthouse report, we would get exponentially better performance gains by deferring the loading of offscreen images. We could definitely defer loading the selectr css though

aslafy-z:
Closing this as the base repository has been deleted. @Larsklopstra Feel free to restore/reopen if you want to finish your work on this one.

PR #252, opened 2019-01-15
Alexa swag campaign is no longer active

etabakov:
- [X] I've checked that this isn't a new swag opportunity proposal.
- [X] I've checked that this isn't a duplicate pull request.

Alexa campaign expired on 31 Dec 2018. Explained here: Explained here: https://developer.amazon.com/alexa-skills-kit/alexa-developer-skill-promotion

### DIFF
--- a/data.json
+++ b/data.json
@@ -1,14 +1,5 @@
 [
   {
-    "name": "Alexa",
-    "difficulty": "hard",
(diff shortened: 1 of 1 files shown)


vikaspotluri123:
Closing in favor of #159

PR #259, opened 2019-01-25
Fix redirect when the `search` param is `false`

itaisteinherz:

This fixes an issue where the `expired` query parameter is changed from `false` to `true`.

See more on this in #159 and https://github.com/swapagarwal/swag-for-dev/pull/159#issuecomment-457642144.

// cc @aslafy-z @vikaspotluri123


aslafy-z:
We can wait for #260 so we're sure that the deployed version works without JS enabled. (#159 still not deployed, dunno why)

PR #260, opened 2019-01-25
Hide expired promos on initial load

itaisteinherz:

This makes sure the expired promos are hidden when loading the website for the first time by hiding all promos by default, and then showing the valid promos using JS.
Note that I tried to do this by using a conditional expression in Pug where the promo gets the `"visible"` class if `swag.expired === false`, but it didn't work for some reason (probably because I don't know how to use Pug).

See more on this in #159 and https://github.com/swapagarwal/swag-for-dev/pull/159#issuecomment-457642144.

// cc @aslafy-z @vikaspotluri123

### DIFF
--- a/src/pug/includes/body.pug
+++ b/src/pug/includes/body.pug
@@ -43,7 +43,7 @@ include ./fork.pug
 				- difficulty = swag.difficulty + ' difficulty';
 				- difficultyTitle = 'Difficulty: ' + swag.difficulty;
 				div(
-					class="item visible " + swag.difficulty + swag.tags.map(t => ` tag-${t}`).join('')
+					class="item " + swag.difficulty + swag.tags.map(t => ` tag-${t}`).join('')
 					data-name=swag.name.toLowerCase()
 					data-difficulty=['easy','medium', 'hard'].indexOf(swag.difficulty)
 				)
--- a/src/pug/includes/body.pug
+++ b/src/pug/includes/body.pug
@@ -44,7 +44,7 @@ include ./fork.pug
 				- difficultyTitle = 'Difficulty: ' + swag.difficulty;
 				- expired = swag.tags.includes('expired');
 				div(
-					class="item " + (!expired ? 'visible ' : '') + swag.difficulty + swag.tags.map(t => ` tag-${t}`).join('')
+					class="item " + (expired ? '' : 'visible ') + swag.difficulty + swag.tags.map(t => ` tag-${t}`).join('')
 					data-name=swag.name.toLowerCase()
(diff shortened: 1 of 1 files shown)


src/pug/includes/body.pug
```diff
@@ -43,7 +43,7 @@ include ./fork.pug
 				- difficulty = swag.difficulty + ' difficulty';
 				- difficultyTitle = 'Difficulty: ' + swag.difficulty;
 				div(
-					class="item visible " + swag.difficulty + swag.tags.map(t => ` tag-${t}`).join('')
+					class="item " + swag.difficulty + swag.tags.map(t => ` tag-${t}`).join('')
```
itaisteinherz: This also works:

```suggestion
					class="item " + (!swag.expired ? ' visible' : '') + swag.difficulty + swag.tags.map(t => ` tag-${t}`).join('')
```

But it seems like it doesn't make the promos load any faster (for me, at least).

itaisteinherz:
@Yash-Garg You don't have to approve this (and other PRs) multiple times. Feel free to comment or suggest anything, but approving PRs multiple times doesn't help anyone.

src/pug/includes/body.pug
vikaspotluri123: I don't think we should worry too much about TTI in this specific case. We have a few other things to improve (most notably how we handle third party scripts) before we should handle stuff like this.

Disabling asset optimization on Netlify's side dropped TTI from 3.7 to 2.7 seconds for me so we're definitely not far from 🔥

itaisteinherz:
@vikaspotluri123 I've added @aslafy-z's suggestion, and so I think this is ready to be merged. Also, I think #259 is ready as well.

src/pug/includes/body.pug
```diff
@@ -42,8 +42,9 @@ include ./fork.pug
 			each swag in swagList
 				- difficulty = swag.difficulty + ' difficulty';
 				- difficultyTitle = 'Difficulty: ' + swag.difficulty;
+				- expired = swag.tags.includes('expired');
 				div(
-					class="item visible " + swag.difficulty + swag.tags.map(t => ` tag-${t}`).join('')
+					class="item " + (!expired ? 'visible ' : '') + swag.difficulty + swag.tags.map(t => ` tag-${t}`).join('')
```
vikaspotluri123:
```suggestion
					class="item " + (expired ? '' : 'visible ') + swag.difficulty + swag.tags.map(t => ` tag-${t}`).join('')
```

aslafy-z:
Ok, it works now with JS disabled. Thanks @itaisteinherz!
Let's merge!

PR #264, opened 2019-02-05
Scale and rotate swag images

### LINKED ISSUE #245: Make images interactive

When I saw the website the first thing I tried to do was a click on an image. How about we made the pictures interactive (and maybe clicking them served the purpose of a link)?

web-man:

Fixes #245 

Added some scale and rotate animations when user hovers item. If you don't like rotate just tell me I will remove it easily.

### DIFF
--- a/src/styl/imports/card.styl
+++ b/src/styl/imports/card.styl
@@ -56,9 +56,16 @@
 	img
 		height 100%
 		width auto
+		transition: all 300ms linear;
 
 	.img-container
 		height 300px
 		width 100%
 		overflow hidden
 		padding-bottom 15px
+
+	&:hover
+		img
+			-ms-transform rotate(3deg) scale(1.05)
+			-webkit-transform rotate(3deg) scale(1.05)
+			transform rotate(3deg) scale(1.05)
--- a/src/styl/imports/card.styl
+++ b/src/styl/imports/card.styl
@@ -56,7 +56,7 @@
 	img
 		height 100%
 		width auto
-		transition: all 300ms linear;
+		transition: all 300ms linear
 
 	.img-container
 		height 300px
@@ -66,6 +66,4 @@
 
 	&:hover
 		img
-			-ms-transform rotate(3deg) scale(1.05)
-			-webkit-transform rotate(3deg) scale(1.05)
 			transform rotate(3deg) scale(1.05)
--- a/src/styl/imports/card.styl
+++ b/src/styl/imports/card.styl
@@ -56,7 +56,7 @@
 	img
 		height 100%
 		width auto
-		transition: all 300ms linear
+		transition all 300ms linear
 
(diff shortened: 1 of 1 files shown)


Yash-Garg:
Would be nice if u add a gif here, so that everyone can see how it looks :)

web-man:
@Yash-Garg Netlify gives us [link](https://deploy-preview-264--peaceful-chandrasekhar-efde8e.netlify.com/) to check changes.

Yash-Garg:
OwO

aslafy-z:
I personaly like it but I think there should be some more random in it. The images are all rotating in the same direction with the same angle. Thoughts?

plibither8:
On mobile, the animation happens when the user _clicks/taps_  on the image rather than _hover_, since that's not possible. The image stays rotated and enlarged until another image is tapped. I think there should be an easier way to "reset" the image to its originals configs, perhaps when tapping/clicking the image again?

Otherwise, I like it :)

src/styl/imports/card.styl
```diff
@@ -56,9 +56,16 @@
...
 
 	.img-container
 		height 300px
 		width 100%
 		overflow hidden
 		padding-bottom 15px
+
+	&:hover
+		img
+			-ms-transform rotate(3deg) scale(1.05)
```
vikaspotluri123: @aslafy-z do we have autoprefixer as part of our build system?

aslafy-z: It seems not, but it could be great to add it.

vikaspotluri123: Created an issue for it 👍

src/styl/imports/card.styl
```diff
@@ -56,9 +56,16 @@
 	img
 		height 100%
 		width auto
+		transition: all 300ms linear;
```
aslafy-z:
```suggestion
		transition: all 300ms linear
```

src/styl/imports/card.styl
```diff
@@ -56,9 +56,16 @@
...
 		height 300px
 		width 100%
 		overflow hidden
 		padding-bottom 15px
+
+	&:hover
+		img
+			-ms-transform rotate(3deg) scale(1.05)
+			-webkit-transform rotate(3deg) scale(1.05)
+			transform rotate(3deg) scale(1.05)
```
aslafy-z: Since we merged #301, you can drop the two vendor specific lines (-ms-transform & -webkit-transform).

swapagarwal:
Looks good to me, can be merged after resolving the conflicts.

web-man:
@swapagarwal Done

swapagarwal:
Thanks @web-man! 👍

web-man:
@swapagarwal You're welcome!

PR #296, opened 2019-07-23
Sort by name when difficulty is the same

thethomasz:

Changed comparator function when sorting by difficulty to sort by name when difficulties are the same.


thethomasz:
Hmm, I can't seem to add reviewers? Should I just leave it?

PR #299, opened 2019-08-17
Adding Google Cloud Opportunity

Xtremilicious:

Adding Google Cloud Study Jam 2019 Opportunity

### DIFF
--- a/data.json
+++ b/data.json
@@ -108,6 +108,15 @@
     "tags": ["clothing", "device"]
   },
   {
+    "name": "Google Cloud",
+    "difficulty": "medium",
+    "description": "Finish 3 Quests on Qwiklabs before the deadline, get recognized by Google Developers, get cool schwags and more! (Ships only to India)",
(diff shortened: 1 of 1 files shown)


aslafy-z:
Well, as it ships only to india, I'm not sure whether we should add this or not. 

We could eventually tag this swag with some label and then filter on it. WDYT @vikaspotluri123 @swapagarwal @plibither8 ?

Yash-Garg:
@aslafy-z August one has ended and they haven't started September registrations yet. So i suggest to wait until they open September registrations

Xtremilicious:
As @Yash-Garg said, if and when September registrations start, it needs to be updated.
As for the India-only-shipping issue, whether to post this or not, depends on other collaborators:)

vikaspotluri123:
Codeship has limited shipping and we've allowed it, so I see no issue with allowing this as well. If the challenge is recurring, I think we should include that in the description and allow the community to submit the updated link, if it changes.

swapagarwal:
This hasn't been updated for September, so closing. These type of items are best shared in the Telegram group! 😉

Issue #297, opened 2019-07-29
Add autoprefixer to build system


vikaspotluri123:
As part of this, we also need to remove prefixed code

_Originally posted by @aslafy-z in https://github.com/swapagarwal/swag-for-dev/pull/264_

plibither8:
Fairly simple, we can use one of the two ways: PostCSS or a direct autoprefixing module. If we use PostCSS, we can use other plugins to optimise CSS output and ease development.

PR #301, opened 2019-08-25
Added autoprefixer to build system

arulprabakaran:

Added autoprifxing css into build system using postcss and autoprefixer (postcss plugin).

Fixes: #297

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -12,6 +12,8 @@ const concat = require('gulp-concat');
 const download = require('gulp-download-stream');
 const responsive = require('gulp-responsive');
 const merge = require('merge-stream');
+const postcss = require('gulp-postcss');
+const autoprefixer = require('autoprefixer');
 
 const {swagList, swagImages} = require('./get-data');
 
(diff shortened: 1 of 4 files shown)


gulpfile.js
```diff
@@ -65,6 +67,9 @@ gulp.task('pug', () => {
 gulp.task('styl', () => {
 	return gulp.src('src/styl/index.styl')
 		.pipe(stylus({compress: true}))
+		.pipe(postcss([
+			autoprefixer({browsers: ['last 1 version']})
+    	]))
```
vikaspotluri123: Looks like there might be some spaces here... or github is messing with me

gulpfile.js
```diff
@@ -65,6 +67,9 @@ gulp.task('pug', () => {
 gulp.task('styl', () => {
 	return gulp.src('src/styl/index.styl')
 		.pipe(stylus({compress: true}))
+		.pipe(postcss([
+			autoprefixer({browsers: ['last 1 version']})
```
vikaspotluri123: I think we should do at least `last 2 versions`

gulpfile.js
arulprabakaran: It is fine in editor, maybe github compare make it look weird.

gulpfile.js
arulprabakaran: Done

PR #335, opened 2019-10-01
Added Blockstack swag

fzs1994:


vikaspotluri123:
This seems to be a more community-oriented swag offering. I think we should focus on swag offerings designed explicitly for individual contributors 🤔

swapagarwal:
It's similar to FSFE, which we ultimately took down. This doesn't help the community as it's bound to get spammed.

Issue #338, opened 2019-10-01
Website: Change default sorting to newest first


swapagarwal:
## Issue regarding the website

- [x] I've checked that this isn't a duplicate issue.

Let's change the default sorting from alphabetical to newest first so that new opportunities are easily discoverable. 🙂

versavale:
Can I try this?

itnug987:
@swapagarwal Can I work on this issue?

vikaspotluri123:
@itnug987 #343 is already in progress

PR #345, opened 2019-10-01
set default sorting to newest first

duydosomething:

Set selected value to true for `Newest First` option so that "Newest First" is selected on page load. Kept current list order in place. Please advise if something else is needed.

This refers to #338


duydosomething:
Ah shit, there's already a PR for this. Closing mine. Sorry!!

PR #362, opened 2019-10-03
Fix Travis build -> Hacktoberfest 2019 image URL

Glennmen:
- [X] I've checked that this isn't a new swag opportunity proposal.
- [X] I've checked that this isn't a duplicate pull request.

This PR fixes the Travis Build error, the current Hacktoberfest 2019 image urls isn't valid anymore.

### DIFF
--- a/data.json
+++ b/data.json
@@ -112,7 +112,7 @@
     "difficulty": "hard",
     "description": "To qualify for the official limited edition Hacktoberfest T-shirt, you must register and then make four pull requests (PRs) between October 1-31 (in any time zone).",
     "reference": "https://hacktoberfest.digitalocean.com/" ,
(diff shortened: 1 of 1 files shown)


Glennmen:
Hmm now throws errors on other items, I would suggest to skip items that are tagged as `expired`.

swapagarwal:
> Hmm now throws errors on other items, I would suggest to skip items that are tagged as `expired`.
> I could try to add this in an other PR.

That makes sense! We should skip checking the items which are tagged as expired.

Glennmen:
@swapagarwal I will make a PR tomorrow for this then, I think especially in the long run images or links to old events can get pulled offline so should trigger a Travis build to fail.

I don't think this fits in the scope of this PR (fixes current event image).

swapagarwal:
Thanks @Glennmen! 👍

PR #365, opened 2019-10-04
Switch to Github Actions

vikaspotluri123:
👀


.github/workflows/main.yml
```diff
@@ -0,0 +1,35 @@
+name: All the CI
+
+on: [push, pull_request]
```
vikaspotluri123: Thinking of changing this to only pushing to master, and PRs - thoughts?

.github/workflows/main.yml
```diff
@@ -0,0 +1,35 @@
...
+
+jobs:
+  unit:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [10.x]
+    steps:
```
vikaspotluri123: Thinking of dropping this. Github has 10.x already installed, so I don't think we need to worry about specifying a node version. This step took a long time in the actions run

swapagarwal:
Why are 2 tests canceled?

vikaspotluri123:
That was the on.push event. We don't need to run it since the pull_request is the same thing

.github/workflows/main.yml
```diff
@@ -0,0 +1,35 @@
...
+    - uses: actions/checkout@v1
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - name: Unit Tests
+      run: |
+        npm ci
+        npm run test
```
vikaspotluri123: Thinking of dropping this. Github has 10.x already installed, so I don't think we need to worry about specifying a node version. This step took a long time in the actions run

_multiline comments!_

.github/workflows/main.yml
```diff
@@ -0,0 +1,35 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [10.x]
+    steps:
+    - uses: actions/checkout@v1
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
```
aslafy-z: You mean this right?

vikaspotluri123: Yep

.github/workflows/main.yml
aslafy-z: For reference, https://help.github.com/en/articles/software-in-virtual-environments-for-github-actions#ubuntu-1804-lts : Node.js (v10.16.3)

vikaspotluri123: Right, and I don't think we're going to run into version issues... ever

vikaspotluri123:
Good to go... and faster than Travis!

.github/workflows/main.yml
aslafy-z: Seems OK, but I don't see the need we have to restrict

vikaspotluri123: I had a reason but I don't anymore 🙃

aslafy-z:
Here's the the badge to add to the readme: https://github.com/swapagarwal/swag-for-dev/workflows/.github/workflows/main.yml/badge.svg
I believe it will work when when this PR will be merged to master.

https://help.github.com/en/articles/configuring-a-workflow#adding-a-workflow-status-badge-to-your-repository

aslafy-z:
If you delete the Travis badge, I guess you can delete the travis config file also

vikaspotluri123:
I was going to hold off on that till we get rid of Travis

swapagarwal:
Let's first merge this and get both Travis and GitHub Actions running the tests. Once the latter is working as expected, we can disable Travis integration.

README.md
```diff
@@ -1,7 +1,7 @@
 # swag-for-dev
 
 [![Awesome](https://awesome.re/badge-flat.svg)](https://awesome.re)
-[![Build Status](https://travis-ci.org/swapagarwal/swag-for-dev.svg?branch=master)](https://travis-ci.org/swapagarwal/swag-for-dev)
```
aslafy-z: i'll go for keeping this line from now as we are not disabling travis right away

vikaspotluri123: done

.github/workflows/main.yml
aslafy-z: You can drop the `pull_request` trigger or it will trigger twice.

.github/workflows/main.yml
```diff
@@ -0,0 +1,21 @@
+name: All the CI
+
+on: [push, pull_request]
+
+jobs:
+  unit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
```
aslafy-z:
```suggestion
    - uses: actions/checkout@v2
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,21 @@
...
+    - name: Unit Tests
+      run: |
+        npm ci
+        npm run test
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
```
aslafy-z:
```suggestion
    - uses: actions/checkout@v2
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,21 @@
+name: All the CI
+
+on: [push, pull_request]
+
+jobs:
+  unit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Unit Tests
```
aslafy-z: Is this well aligned?

.github/workflows/main.yml
vikaspotluri123: It will only trigger for branches from this repo that are PR'd, correct? E.g. if you create a PR from your fork, CI would never run. I think that might have been why I only wanted CI to run on master + PRs

.github/workflows/main.yml
vikaspotluri123: Seems so, CI runs properly

vikaspotluri123: Yep, the steps in CI are correct 😄

vikaspotluri123:
CI found an actual issue, but it actually runs now

.github/workflows/main.yml
aslafy-z: My app was showing an extra space. Sorry for that.

.github/workflows/main.yml
aslafy-z: You are right ! Let's re-add the master specification for push event.

vikaspotluri123:
Everything looks good now!

.github/workflows/main.yml
```diff
@@ -1,7 +1,12 @@
 name: All the CI
 
-on: [push, pull_request]
-
+on:
+  push:
+    branches:
+      - master
+  pull_request:
```
aslafy-z: Can be shortened to pull_request: {}

vikaspotluri123: Just curious, do you have any links to the docs for that? I looked everywhere for them on a different project but couldn't find 'em!

aslafy-z: Nothing but https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#on
It is implicit, it should get the same result as
```
on:
  - pull_requests
```
Or, to be precise you can do :
```
on:
  pull_request:
    types:
      - created
      - synchronize
```

vikaspotluri123:
@swapagarwal I think this is ready to be merged?

.github/workflows/main.yml
aslafy-z: In fact, could you please specify events like I did in my second sample?

aslafy-z:
LGTM! 😊 
This should be merged!

.github/workflows/main.yml
vikaspotluri123: Do we want to use that exact list, or anything additionally?

aslafy-z: I was almost sure I did not pushed the comment button in here :o.
I did some tests by renaming the PR (trigger the `edited` event), actions were not re-triggered, so I guess the defaults for `pull_request` already excludes theses configs and thus already fits our need!

vikaspotluri123:
Just waiting for @swapagarwal 👋

aslafy-z:
Let's merge that one so we can get rid of travis! @swapagarwal any comment?

swapagarwal:
Go ahead!

PR #374, opened 2019-10-04
Prevent webpage loading jank

### LINKED ISSUE #366: Reduce website loading jank
With #343, we updated the default sort order of the swag - from alphabetical to newest first. We didn't update the order that the swag is listed at build time. This leads to jank:

jank

The solution to this is to update the `get-data` script we use to return the swag list, sorted by date.

https://github.com/swapagarwal/swag-for-dev/blob/9e4df0b188e460f26df81cfb7a32e2fa5ab2f613/get-data.js#L3-L26

Jeremy-Mitchell89:
- [x ] I've checked that this isn't a new swag opportunity proposal.
- [x ] I've checked that this isn't a duplicate pull request.

added a few lines of code to close #366, sorts the swag list by date to prevent the loading jank

### DIFF
--- a/get-data.js
+++ b/get-data.js
@@ -24,5 +24,10 @@ const swagList = require('./data.json').map(swag => {
 
 	return swag;
 });
-
+//Sort list by most recent
+swagList.sort(function (a, b) {
+	a = new Date(a.dateAdded);
+	b = new Date(b.dateAdded);
+	return a > b ? -1 : a < b ? 1 : 0;
+});
 module.exports = {swagList, swagImages};


get-data.js
```diff
@@ -24,5 +24,10 @@ const swagList = require('./data.json').map(swag => {
 
 	return swag;
 });
-
+//Sort list by most recent
+swagList.sort(function (a, b) {
+	a = new Date(a.dateAdded);
+	b = new Date(b.dateAdded);
+	return a > b ? -1 : a < b ? 1 : 0;
+});
```
vikaspotluri123:
```suggestion
// Sort by newest opportunities
}).sort(({dateAdded: left}, {dateAdded: right}) => new Date(right) - new Date(left));
```

vikaspotluri123: I'm not sure if comitting this suggestion will mess up the file... Multiline comments is in beta, so this might be completely broken 😄

aslafy-z:
We need that one merged!

vikaspotluri123:
@Jeremy-Mitchell89 ping! We have another contributor looking to tackle this, if you don't have time to finish this, please let us know

PR #397, opened 2019-10-07
Update data.spec.js

vasu-gupta:
I've checked that this isn't a duplicate pull request.
The change I made in the data.spec.js file is under the checkURL function for the condition of false as for when in false state it should be 'got' instead of 'get'.
**function checkURL(url, head = false) {
	const method = head ? 'head' : 'get'; //'got'**
#swag-for-dev
Thank you!

### DIFF
--- a/test/functional/data.spec.js
+++ b/test/functional/data.spec.js
@@ -4,7 +4,7 @@ const {expect} = require('chai');
 const data = require('../../data.json');
 
 function checkURL(url, head = false) {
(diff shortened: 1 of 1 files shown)


aslafy-z:
Hello @vasu-gupta! Can you explain why it should be got instead of get? Thanks

vikaspotluri123:
I'm going to close this for now, since CI is also failing. If we get an explanation for why this is needed, I'm more than happy to reopen it.

Issue #393, opened 2019-10-05
Difficulty label gets distorted with long titles 


aslafy-z:
## Issue regarding the website

- [x] I've checked that this isn't a duplicate issue.

Difficulty indicator gets distorted on mobile with long titles :

![Screenshot_20191006_014007_com android chrome](https://user-images.githubusercontent.com/8191198/66262144-51ef4480-e7db-11e9-8b7a-6d0e5689b9ba.jpg)

Chime on Android

mpaarating:
I'll start working on this.

aslafy-z:
Fixed by #401

PR #401, opened 2019-10-07
Fix responsive sizing for difficulty tag

mpaarating:

## Description
Per issue #393, the difficulty label distorts at low screen width sizes.

## How to test
Run the project and resize the window to the desired screen size to confirm consistent behavior.

## Screenshot
[Image of responsive change, resolution 282 x 648](https://gyazo.com/a347803da3a5be45633103b47d413a19)

### DIFF
--- a/src/pug/includes/body.pug
+++ b/src/pug/includes/body.pug
@@ -53,7 +53,8 @@ include ./fork.pug
 				)
 					div.title.flex
 						h1=swag.name
-						div(class=difficulty title=difficultyTitle)
+						div
+							div(class=difficulty title=difficultyTitle)
 							span.sr-only=difficultyTitle
 					p.swag
 						each swagTag in swag.tags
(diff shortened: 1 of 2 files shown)


src/pug/includes/body.pug
```diff
@@ -53,7 +53,8 @@ include ./fork.pug
 				)
 					div.title.flex
 						h1=swag.name
-						div(class=difficulty title=difficultyTitle)
+						div
+							div(class=difficulty title=difficultyTitle)
 							span.sr-only=difficultyTitle
```
vikaspotluri123:
```suggestion
								span.sr-only=difficultyTitle
```

vikaspotluri123: - The span should be nested in the div since it's labeling it

mpaarating:
@vikaspotluri123 Good catch, thank you!

mpaarating:
Looks like one of the `Twilio` entries is causing the tests to fail. When I run it locally, that `Twilio` entry passes, but the `LBRY` one fails.

PR #415, opened 2019-10-09
Parallel tests

### LINKED ISSUE #385: Run tests faster

Right now, it takes around 2-3 minutes to run all the tests. Brainstorm on how can we achieve the same in seconds.

driver-by:

Adding parallel tests with mocha.parallel
Fixes #385

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2916,7 +2916,8 @@
     "deep-extend": {
       "version": "0.6.0",
       "resolved": "https://registry.npmjs.org/deep-extend/-/deep-extend-0.6.0.tgz",
-      "integrity": "sha512-LOHxIOaPYdHlJRtCQfDIVZtfw/ufM8+rVj649RIHzcm/vGwQRXFt6OPqIFWsm2XEMrNIEtWR64sY1LEKD2vAOA=="
+      "integrity": "sha512-LOHxIOaPYdHlJRtCQfDIVZtfw/ufM8+rVj649RIHzcm/vGwQRXFt6OPqIFWsm2XEMrNIEtWR64sY1LEKD2vAOA==",
+      "dev": true
     },
     "deep-is": {
       "version": "0.1.3",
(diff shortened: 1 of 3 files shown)


vikaspotluri123:
Wow, mocha.parallel seems like what we were looking for!

I think we might want to split it so we limit the number of requests that are made at once.

swapagarwal:
LGTM!
@vikaspotluri123 The domains are all different so why would we want to limit the number of simultaneous requests?

vikaspotluri123:
Outbound traffic might be throttled to prevent abuse, and since the number of items is not static, we don't know exactly how many requests we're going to be sending. If you look at the log, the test times are progressively longer, unlike the current state, probably because of various factors such as that

swapagarwal:
If the `checkURL` function was asynchronous, then we might have that problem but since it's not, the number of parallel tests would be determined by the number of cores, no?

driver-by:
Don't sure if there could be reasonable limitation of outbound requests, as loading modern website causes 50-100 parallel requests to happen.
Speaking of growing database, do you plan to remove old data?
If not, then 200+ simultaneous requests might be not a good idea :)
As of limitation of parallelization, it should be limited to number of threads that could be created within one process. Fot my linux machine it's
```
cat /proc/sys/kernel/threads-max
125029
```

driver-by:
I've tried a code
```
const LIMIT_PARALLEL_TESTS = 40;
//.....

	let dataSlices = [];
	for (let i = 0, length = data.length; i < length; i += LIMIT_PARALLEL_TESTS) {
		dataSlices.push(data.slice(i, i + LIMIT_PARALLEL_TESTS));
	}
	dataSlices.forEach(data => {
		parallel('validate images and references', function () {
			data.forEach(opportunity => {
				//...
			});
		});
	});
//...
```
(moved `expect(data).to.be.an('Array');` to `"data.json is valid"` test)
And it works as intended.
Let me know if you want me to commit it.

vikaspotluri123:
> If the `checkURL` function was asynchronous, then we might have that problem but since it's not, the number of parallel tests would be determined by the number of cores, no?

checkURL is async, is it not?



> Let me know if you want me to commit it.
@driver-by go ahead and commit it, if we run into an issue, it's easy to revert it 😄

driver-by:
> @driver-by go ahead and commit it, if we run into an issue, it's easy to revert it smile

Right, added a commit :D

>checkURL is async, is it not?

Getting into your conversation again, `checkURL` is async as it performs async HTTP call and returns a promise.

vikaspotluri123:
I think splitting it up into 10 requests is good... a lot fewer requests are yellow or redlining.

aslafy-z:
Merging that one, thank you @driver-by!

Issue #435, opened 2019-10-15
Operation Code marked 'EASY'


KillMonger1:
## Issue with an existing swag opportunity

- [x] I've checked that this isn't a duplicate issue.

Q. Which opportunity?  
A. : Operation Code

The Operation Code Swag opportunity is marked as easy. I feel that resolving 3 issues or merging 2 PR's aren't less than 'medium' difficulty for any newbie.

wuworkshop:
Just submitted a quick PR for this one.

PR #444, opened 2019-10-18
Close: #435 - Change Operation Code difficulty level to 'medium'.

wuworkshop:

This resolves issue #435 - changing Operation Code's difficulty level from 'easy' to 'medium'.

### DIFF
--- a/data.json
+++ b/data.json
@@ -317,9 +317,9 @@
   {
     "name": "Operation Code",
     "difficulty": "medium",
-    "description": "Close 3 issues, get a t-shirt. Make 2 PRs and get stickers.",
+    "description": "Resolve 3 issues, get a t-shirt. Merge 2 PRs and get stickers.",
     "reference": "https://github.com/OperationCode/START_HERE",
-    "image": "https://upload.cc/i1/2018/10/21/yRDFvh.png",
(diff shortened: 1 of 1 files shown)


swapagarwal:
Thanks @wuworkshop! 👍

data.json
```diff
@@ -316,10 +316,10 @@
   },
   {
     "name": "Operation Code",
-    "difficulty": "easy",
-    "description": "Close 3 issues, get a t-shirt. Make 2 PRs and get stickers.",
+    "difficulty": "medium",
+    "description": "Resolve 3 issues, get a t-shirt. Merge 2 PRs and get stickers.",
     "reference": "https://github.com/OperationCode/START_HERE",
```
aslafy-z: I guess we can change that image but we would need to rehost it on imgur for example.

data.json
```diff
@@ -75,7 +75,7 @@
     "name": "CircleCI",
     "difficulty": "medium",
     "description": "Participate in #Orbtoberfest with CircleCI and claim limited edition swag! 1 - 3 pull requests: Limited-Edition Sticker, 4+ pull requests: Limited-Edition T-shirt",
-    "reference": "https://hacktoberfest.circleci.com/#/" ,
+    "reference": "https://hacktoberfest.circleci.com/#/",
```
aslafy-z: Let's correct this url at the same time. Please drop the trailing #/

data.json
swapagarwal: Sorry I merged early. 🙈

PR #455, opened 2019-10-23
Fix image generation

Mox10:
- [ X ] I've checked that this isn't a new swag opportunity proposal.
- [ X ] I've checked that this isn't a duplicate pull request.

I have changed the img width in card.styl to be 100% instead of auto. This fixes the image displayed on the site so they are no longer zoomed in.


aslafy-z:
IMHO this should stay like it is now. It looks better. WDYT @vikaspotluri123 @swapagarwal?

vikaspotluri123:
I agree, it looks more uniform

aslafy-z:
@Mox10 we'll stay with what currently have. Thank you anyway!

PR #462, opened 2019-10-28
ReadMe updated

vasu1818:


aslafy-z:
Hello @vasu1818. Sorry but we cannot accept this kind of pull request. You can understand that it doesn't add any value to the project, it looks like a spam PR. See https://dev.to/mpermar/how-to-create-effective-pull-requests-2m8e

Issue #447, opened 2019-10-18
Fix image generation 


aslafy-z:
This upgrade changed the way images are generated in some way. The generated images aren't the same. We might need to adjust some settings.

_Originally posted by @aslafy-z in https://github.com/swapagarwal/swag-for-dev/pull/440#issuecomment-543343226_

ajitsinghkaler:
@aslafy-z Can I work on this.

aslafy-z:
Nice, I assigned it to you. Have a look to the linked description :arrow_up:

ajitsinghkaler:
I need to know the change we need to make to the images. Can you please attach some screenshots. I downgraded the gulp-responsive version but the images were still the same.

ajitsinghkaler:
@aslafy-z This is not an image generation issue most of it occurs because most images are not 300*300 so for that either we can change the CSS or reduce the size of images to width:300 px. I'm making both changes please review and tell which is more appropriate.

ajitsinghkaler:
What you think is an appropriate solution I can create a pull request for that.

PR #443, opened 2019-10-17
feat: redesign website

anhek:
Hello :)

I have finished the redesign. There's still some things to do like : 

- adding the colours on the tags
- customizing the form

I did some research and it looks like you need to have a paid subscription on Upscribe to have the ability to add custom theme :/

I'd like someone to help me getting the cards to have the same height like on this article please : https://www.lottejackson.com/learning/an-equal-height-grid-using-flexbox

equal-height-grid

Feedback time ! 🙂

PR #466, opened 2019-10-30
Image generation bug gulp

ajitsinghkaler:

Images were first made responsive according to height but according to the CSS applied we have to set it according to the width. Which is what I've done I've set it according to width.

### DIFF
--- a/data.json
+++ b/data.json
@@ -99,6 +99,15 @@
     "tags": ["clothing", "expired"]
   },
   {
+    "name": "Datenanfragen.de e. V.",
+    "difficulty": "medium",
+    "description": "One merged PR gets at least 3 stickers (limited to 100 sets) and a chance at a limited edition t-shirt.",
+    "reference": "https://www.datarequests.org/blog/hacktoberfest-2019/",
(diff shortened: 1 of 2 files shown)


aslafy-z:
Thank you @ajitsinghkaler! 
I'm not sure it will fit the redesign done by @anhek in #443. Can you have a look?

ajitsinghkaler:
Yeah, it will look really bad, so what can we do with images which don't fit into our aspect ratio should I enlarge them or what because cramping existing images into that section is not working. need your suggestion on this.


Desktop screenshot (12)

ajitsinghkaler:
So for that, we have to make no changes and we can keep our code as it is and close issue number #447

ajitsinghkaler:
but then again the third image is cut down because this has it is not at a specific aspect ratio.

ajitsinghkaler:
Cn you please look into this

ajitsinghkaler:
@swapagarwal @aslafy-z @vikaspotluri123 Can you please tell me what I should do with this

PR #469, opened 2019-11-04
Added expired tag to hacktoberfest issues

### LINKED ISSUE #468: Remove HacktoberFest related Swags
## Issue with an existing swag opportunity

Q. Which opportunity?  
A. : Many 

There are many Hacktoberfest related Swag opportunities which seem to have expired as HacktoberFest has ended. Why not remove all those.

KillMonger1:

I've added an expired tag to all hacktoberfest issues after checking. I've left out the docusaurus opportunity as i think it is still valid.
Fixes #468


aslafy-z:
Done with #473. Thanks 👌🏼

PR #470, opened 2019-11-04
Update .gitignore for MacOs users

helios1101:

Updated the .gitignore file so that macOS users don't have to manually exclude the unnecessary files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,3 +81,6 @@ typings/
 
 
 # End of https://www.gitignore.io/api/node
+
+## macOS 
+.DS_Store


aslafy-z:
Thanks for the proposal @helios1101.
I vote for using this directly: https://www.gitignore.io/api/node,macos

aslafy-z:
Finally, we'll not merge that one. You may want to create a `.gitignore` file in any parent directory as explained in here: https://git-scm.com/docs/gitignore.
Let keep this gitignore specialized, or we will start to add others oses, editors ... entries. Thanks again !

PR #565, opened 2020-04-12
Change image of erxes because current image gives 404

SemperDisco:
Changed the image of the erxes opportunity because the current image gives a 404 in the test.

### DIFF
--- a/data.json
+++ b/data.json
@@ -166,7 +166,7 @@
     "difficulty": "medium",
     "description": "Submit 3 issues and 1 PR to their <a href='https://github.com/erxes/erxes'>repo</a>, get a t-shirt and stickers.",
     "reference": "https://erxes.io/hubspot-alternative-erxes-swag",
-    "image": "https://erxes.io/static/images/t-shirt.jpeg",
+    "image": "https://i.imgur.com/pE9Ml3v.png",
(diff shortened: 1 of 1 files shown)


data.json
```diff
@@ -166,7 +166,7 @@
     "difficulty": "medium",
     "description": "Submit 3 issues and 1 PR to their <a href='https://github.com/erxes/erxes'>repo</a>, get a t-shirt and stickers.",
     "reference": "https://erxes.io/hubspot-alternative-erxes-swag",
-    "image": "https://erxes.io/static/images/t-shirt.jpeg",
+    "image": "https://i.imgur.com/T1G9zWs.png",
```
vikaspotluri123, line 169:
```suggestion
    "image": "https://i.imgur.com/pE9Ml3v.png",
```

SemperDisco: Thanks you for the review! I changed the url and amended the change.

SemperDisco:
@vikaspotluri123 Could it be that my amend broke the deploy preview?

aslafy-z:
Was due to some connection reset during image download.
Force the build to run again did the trick.

aslafy-z:
Thank you @SemperDisco :100:

PR #580, opened 2020-05-02
feat(ci): install nodejs version according to nvmrc

aslafy-z:
https://github.com/actions/setup-node/issues/32#issuecomment-525791142


aslafy-z:
@vikaspotluri123 can you have a look? I removed the `name`s so we can clearly see what's going on in these steps. No choice but duplicating the node setup code sadly (https://github.community/t5/GitHub-Actions/Support-for-YAML-anchors/m-p/30336).

PR #589, opened 2020-05-04
lazy-load and inline css

### LINKED ISSUE #586: Lazy load CSS resources
Opened after #250 closure.

ritesh-pandey:
Implemented lazy loading of CSS

Instead of `href` attribute, define a `data-href` attribute in `link` tag.
When the `DOMContentLoaded` event is fired, lazy loading method will trigger.
It will look for `link[data-href]` elements in the document and create `href`
attribute with its value set to `data-href`

Fixes #586

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -195,7 +195,9 @@ gulp.task('watch', () => {
 
 gulp.task('inlinesource', () => {
 	return gulp.src('dist/*.html')
-		.pipe(inlinesource())
+		.pipe(inlinesource({
+			rootpath: './dist/'
+		}))
 		.pipe(gulp.dest('dist/', {
 			overwrite: true
 		}));
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -14,7 +14,7 @@ const responsive = require('gulp-responsive');
 const merge = require('merge-stream');
 const postcss = require('gulp-postcss');
 const autoprefixer = require('autoprefixer');
-const inlinesource = require('gulp-inline-source');
+const inlinesource = require('@exuanbo/gulp-inline-source');
 
 const {swagList, swagImages} = require('./get-data');
 
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -34,15 +34,15 @@ let manifest = {
 	'js/index.js': 'js/index.js'
 };
 
-const assetsToBeInlined = ['css/index.css', 'js/index.js'];
+const assetsToBeInlined = ['css/index.css'];
 
-async function getInlinedAssets(config) {
+async function getInlinedAssets(config, manifest) {
 	const fileMap = new Map();
 	const promises = [];
 	const assetBasePath = 'dist/assets/';
 
 	config.forEach(name => {
-		const assetPath = assetBasePath + name;
+		const assetPath = assetBasePath + manifest[name];
 		promises.push(readFile(assetPath, 'utf8')
 			.then(contents => fileMap.set(name, contents)));
 	});
@@ -70,7 +70,7 @@ gulp.task('pug', async done => {
 		js: `/assets/${manifest['js/index.js']}`
 	};
 
-	const inlinedAssets = await getInlinedAssets(assetsToBeInlined);
+	const inlinedAssets = await getInlinedAssets(assetsToBeInlined, manifest);
 
 	gulp.src('src/pug/*.pug')
 		.pipe(pug({
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -206,7 +206,7 @@ gulp.task('inlinesource', () => {
 gulp.task('build', gulp.series(
 	'clean',
 	gulp.parallel(
-		gulp.series('swag-img', 'cachebust', 'styl', 'pug', 'inlinesource'), 'js', 'binaries'
+		gulp.series('swag-img', 'styl', 'cachebust', 'pug', 'inlinesource'), 'js', 'binaries'
 	)
 ));
 
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -14,7 +14,7 @@ const responsive = require('gulp-responsive');
 const merge = require('merge-stream');
 const postcss = require('gulp-postcss');
 const autoprefixer = require('autoprefixer');
-const inlinesource = require('@exuanbo/gulp-inline-source');
+const {readFile} = require('fs').promises;
 
 const {swagList, swagImages} = require('./get-data');
 
@@ -34,7 +34,24 @@ let manifest = {
 	'js/index.js': 'js/index.js'
 };
 
-gulp.task('pug', () => {
+const assetsToBeInlined = ['css/index.css', 'js/index.js'];
+
+async function getInlinedAssets(config) {
+	const fileMap = new Map();
+	const promises = [];
+	const assetBasePath = 'dist/assets/';
+
+	config.forEach(name => {
+		const assetPath = assetBasePath + name;
+		promises.push(readFile(assetPath, 'utf8')
+			.then(contents => fileMap.set(name, contents)));
+	});
+
+	await Promise.all(promises);
+	return fileMap;
+}
+
+gulp.task('pug', async done => {
 	const tags = Array.from(swagList.reduce(
 		(tagList, {tags}) => {
 			tags.filter(tag => tag !== 'expired').forEach(tag => tagList.add(tag));
@@ -53,16 +70,19 @@ gulp.task('pug', () => {
 		js: `/assets/${manifest['js/index.js']}`
 	};
 
-	return gulp.src('src/pug/*.pug')
+	const inlinedAssets = await getInlinedAssets(assetsToBeInlined);
+
+	gulp.src('src/pug/*.pug')
 		.pipe(pug({
 			pretty: true,
-			locals: {swagList, tags, bustedAssets}
+			locals: {swagList, tags, bustedAssets, inlinedAssets}
 		}))
 		.pipe(htmlmin({
 			collapseWhitespace: true,
 			removeComments: true
 		}))
-		.pipe(gulp.dest('dist/'));
+		.pipe(gulp.dest('dist/'))
+		.on('end', () => done());
 });
 
 gulp.task('styl', () => {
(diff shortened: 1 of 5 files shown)


vikaspotluri123:
I think lazy-loading css is a good idea, but given we have a very small amount (9.6k uncompressed), we might find it useful to just inline the entire sheet. I definitely think we should lazy-load the selectr third party dep though 😄

aslafy-z:
OK to inline our CSS and we should definitely lazy-load Selectr. Could you adapt your code to only lazy-load that library?

Thank you!

src/pug/includes/head.pug
```diff
@@ -10,8 +10,8 @@ meta(name="description" content="swag opportunities for developers")
 link(rel="icon" type="image/png" href="/assets/img/logo.png")
 link(rel="apple-touch-icon" size="128x128" href="/assets/img/logo.png")
 
-link(rel="stylesheet" href=bustedAssets.css)
-link(rel="stylesheet" href="https://cdn.jsdelivr.net/npm/mobius1-selectr@2.4.13/dist/selectr.min.css")
+link(rel="stylesheet" data-href=bustedAssets.css)
+link(rel="stylesheet" data-href="https://cdn.jsdelivr.net/npm/mobius1-selectr@2.4.13/dist/selectr.min.css")
```
aslafy-z, line 14: This entry should be duplicated into a `noscript` tag so the style can load even without JS enabled.

src/pug/includes/head.pug
```diff
@@ -10,9 +10,11 @@ meta(name="description" content="swag opportunities for developers")
...
-link(rel="stylesheet" href=bustedAssets.css)
-link(rel="stylesheet" href="https://cdn.jsdelivr.net/npm/mobius1-selectr@2.4.13/dist/selectr.min.css")
+link(rel="stylesheet" inline=true href=bustedAssets.css)
+link(rel="stylesheet" data-href="https://cdn.jsdelivr.net/npm/mobius1-selectr@2.4.13/dist/selectr.min.css")
 
 noscript
 	style.
 		.buttons {display: none}
+
+	link(rel="stylesheet" href="https://cdn.jsdelivr.net/npm/mobius1-selectr@2.4.13/dist/selectr.min.css")
```
aslafy-z, line 20: In fact, I've just seen that we hide the "buttons", including Selectr on noscript. So this can be dropped from here.

gulpfile.js
```diff
@@ -14,6 +14,7 @@ const responsive = require('gulp-responsive');
 const merge = require('merge-stream');
 const postcss = require('gulp-postcss');
 const autoprefixer = require('autoprefixer');
+const inlinesource = require('gulp-inline-source');
```
aslafy-z, line 17: I think we should move to https://github.com/exuanbo/gulp-inline-source since https://github.com/fmal/gulp-inline-source looks to be no more maintained for few years now

ritesh-pandey: I browsed the most used plugin in `gulp` and found this version. Did not look at its maintenance status. Will change to actively maintained version.

aslafy-z:
Looks like it does not build.. Any idea why?

ritesh-pandey:
The production build is failing. Looks like production build process is different than development.

ritesh-pandey:
It is because of race condition between different `gulp` tasks. `styl`, `js`, `swag-img` should be completed before `cachebust` begins. We have not enforced this in code. In current setup, `styl`, `js`, `swag-img` run in parallel. We only enforce that `cachebust` runs after `swag-img`

I think we should enforce this.

```
gulp.series(
'clean',
gulp.parallel(gulp.series('swag-img', 'styl', 'js', 'cachebust', 'pug', 'inlinesource'), 'binaries')
```

aslafy-z:
Well spoted! 

Before: https://lighthouse-dot-webdotdevsite.appspot.com//lh/html?url=https%3A%2F%2Fdevswag.io
After: https://lighthouse-dot-webdotdevsite.appspot.com//lh/html?url=https%3A%2F%2Fdeploy-preview-589--peaceful-chandrasekhar-efde8e.netlify.app%2F

I've seen something weird behavior where it does not shows the background initially:
image

Any thoughts on that?

ritesh-pandey:
We use a SVG file (tiny-grid.svg) as background in hero section. This file needs to be downloaded by the browser before it is rendered.

When rendering a page, browser puts all the assets (images, fonts, scripts, etc.) in a network queue for downloading. In this case, our background SVG is getting pushed at bottom in the network request queue. So we see the page without background image for a while before the downloaded SVG takes its place.

There is a contradictory thing happening here. After inlining the CSS, the rule which says to download the SVG file occurs at very beginning in the file. Compare this to the `img` tags which say to download swag images in the file. Still the swag images are getting placed higher in the download queue.

This may be related to internal working of browser. CSS parser might be slower as compared to HTML and hence the difference in queuing order. Just a guess.


Possible fix
Backgrounds can be given in multiple layers. We can give solid blue color which resembles close to the SVG as a fallback background. When the SVG download completes, solid color will automatically get replaced by SVG.

```CSS
               background url('/assets/img/tiny-grid.svg') repeat, #00a4fb
```

aslafy-z:
We could probably inline it anyway since it's pretty small. In fact this is out of this PR scope. I think we're good to go! @vikaspotluri123 anything to add?

vikaspotluri123:
Since we're already using asset busting, and we only have one set of local styles, I'd rather read the styles during the pug stage since gulp supports Promises. Thoughts?

ritesh-pandey:
I am not aware of Promise based approach. I can look into it.

I suggest that we take configuration based approach. In future if we want to inline any other asset, it should be possible by adding a config. It should be independent of our build script.

I agree to the point of not adding any dependency if the work can be achieved in few lines of code. But at the same time reinventing the wheel should be avoided.

vikaspotluri123:
> I agree to the point of not adding any dependency if the work can be achieved in few lines of code. But at the same time reinventing the wheel should be avoided.

To be clear, my thinking is to read in all the files and pass it in to the pug config, like we pass the `bustedAssets` object. Something like this:

```js
// gulpfile
const inlinedAssets = await Promise.all(readInlinedFiles);

// in gulpfile.task('pug')
.pipe(pug({
			pretty: true,
			locals: {swagList, tags, bustedAssets, inlinedAssets}
		}))
```

```pug
// head.pug. I don't use pug often so the syntax might be a bit off
style
	!=inlinedAssets.css
```

ritesh-pandey:
Hi @vikaspotluri123 
I am kind of stuck and need some help.

```Javascript
// gulpfile

const fs = require('fs');
const { promisify } = require('util');

const readFileAsync = promisify(fs.readFile);

const assetsToBeInlined = {
	'css/index.css': 'dist/assets/css/index.css',
	'js/index.js': 'dist/assets/js/index.js' // taking an example case of multiple assets
};

async function getInlinedAssets(config) {
	let fileMap = new Map();
	for (let [name, path] of Object.entries(config)) {
		fileMap.set(`${name}`, readFileAsync(`${path}`, 'utf8'));
	}
	const result = await Promise.all(fileMap);
	return result;
}
```

Inside `pug` task, I am unable to get resolved value of promise which I have to pass as locals.
Output is shown in comments.

```js
// pug task
const inlinedAssets = getInlinedAssets(assetsToBeInlined);
console.log(inlinedAssets);  // Promise { <pending> }
	.pipe(pug({
			pretty: true,
			locals: {swagList, tags, bustedAssets, inlinedAssets}
		}))
	
```

The value that I receive in `pug` file.
```pug
// pug file
- console.log(inlinedAssets)
//
//Promise {
//  [ [ 'css/index.css', [Promise] ], [ 'js/index.js', [Promise] ] ]
//}


style
	!=inlinedAssets['css/index.css']

```

vikaspotluri123:
I've modified your code and left some comments! Please let me know if anything doesn't make sense 😄 

```js
// Since we're using node 12.x, the fs promises api is no longer experimental
const {readFile} = require('fs').promises;
// You don't necessarily need to use an object here, you could use an array.
// To get the path, just take the asset and add the dist prefix
// e.g. const assets = ['css/index.css', 'js/index.js']
// assets.forEach(asset => console.log({asset, dist:  `dist/assets/${asset}`}));
const assetsToBeInlined = {
	'css/index.css': 'dist/assets/css/index.css',
	'js/index.js': 'dist/assets/js/index.js'
};

async function getInlinedAssets(config) {
	const fileMap = new Map();
	const promises = [];
	for (const [name, path] of Object.entries(config)) {
		promises.push(
			// Your code stored a list of Promises as the value
			// That means if you want to get the value, you have to use `.then`
			// e.g. fileMap.get(name).then(/* logic goes here */);
			// What this does is read the file contents, and after the contents are
			// read, store the contents in the map. Since this is an async function,
			// it's added to a list of promises
			readFile(path, 'utf8').then(contents => fileMap.set(name, contents)
		)
	}
	// fileMap is a Map, not an array. Promise.all expects an array. I'm not sure of its behavior when it gets a Map
	await Promise.all(promises);
	return fileMap;
}


// Pug:
// Funcation signature needs to be updated to `async done => {` since we use await
const inlinedAssets = await getInlinedAssets(assetsToBeInlined);

// No need to return since we're using a callback
gulp.src(...)
	.pipe(pug({
			pretty: true,
			locals: {swagList, tags, bustedAssets, inlinedAssets}
		}))
// gulp.src emits an `end` event when the streaming is complete. Listen for this and signal task completion by calling `done`
    .on('end', () => done())
```

aslafy-z:
@ritesh-pandey Thanks for your changes. Please reset package-lock.json to it's initial state. I think we'll be good then. @vikaspotluri merge when you're ready.

ritesh-pandey:
In production mode, cache busting changes the file name of assets. I need to accommodate that in inlining task.

src/pug/includes/head.pug
```diff
@@ -10,8 +10,10 @@ meta(name="description" content="swag opportunities for developers")
 link(rel="icon" type="image/png" href="/assets/img/logo.png")
 link(rel="apple-touch-icon" size="128x128" href="/assets/img/logo.png")
 
-link(rel="stylesheet" href=bustedAssets.css)
-link(rel="stylesheet" href="https://cdn.jsdelivr.net/npm/mobius1-selectr@2.4.13/dist/selectr.min.css")
+style
+	!=inlinedAssets.get('css/index.css')
+
+link(rel="stylesheet" data-href="https://cdn.jsdelivr.net/npm/mobius1-selectr@2.4.13/dist/selectr.min.css")
```
vikaspotluri123, line 16: I think this still needs a noscript version

ritesh-pandey: `Selectr` is used in buttons section of page. We hide that section if JS is disabled. So IMO loading CSS file in noscript version doesn't make sense.

Please let me know if I missed anything.

vikaspotluri123: Aah, you're right

vikaspotluri123:
@ritesh-pandey it took a bit of time because one of the sites was dead, but your change is now live on https://devswag.io 🎉

PR #595, opened 2020-05-07
fix(test): retry requests 3 times

aslafy-z:
I can see lot of resources tests timing out. Let's retry requesting 3 times before marking the test as failed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "start": "gulp",
     "lint": "eclint check && xo",
     "lint:fix": "eclint fix && xo --fix",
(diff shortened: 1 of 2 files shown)


vikaspotluri123:
If your method doesn't work, I'm assuming the CLI has a retries flag. We can probably use that since the only other tests are basic validation tests that don't take long

aslafy-z:
It should work actually: https://mochajs.org/api/mocha#retries
Maybe it's not due to mocha parallel

vikaspotluri123:
I agree it _should_ work, but I was suggesting this as the alternative since it's not 😄

PR #597, opened 2020-05-07
fix(lint): stay consistent with quotes

aslafy-z:
- pug: single quotes for js and double for html
- stylus: single quotes for pure css, double for stylus (eg. Imports)
- js: single and magic quotes everywhere

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -19,14 +19,14 @@
 [[headers]]
     for = "/assets/swag-img/*"
     [headers.values]
-    Cache-Control = 'public max-age=63072000' # 2 years cache for all assets since they're cachebusted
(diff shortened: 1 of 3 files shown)


vikaspotluri123:
Single quotes in JS, double quotes everywhere else?

aslafy-z:
And single quotes in CSS too (excluding stylus imports), not really sure about that one. That's what I've found to be used almost every where in here. XO has single quotes default for JS, I've found that Google recommends doubles for HTML and simple for css. Wdyt?

vikaspotluri123:
That's about the same with me, though I tend to use single quotes more out of habit. I think your choice is fine.

PR #601, opened 2020-05-13
chore(ci): fix npm cache path not being expanded

aslafy-z:
See https://github.com/swapagarwal/swag-for-dev/runs/668199195

image

`$HOME` is not expanded to `/home/runner` which causes cache Post Run action to fail.
Changing it to `~` as seen in https://github.com/actions/cache/blob/master/examples.md#node---npm fixes the issue.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,7 +14,7 @@ jobs:
 
     - uses: actions/cache@v1
       with:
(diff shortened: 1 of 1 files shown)


vikaspotluri123:
Cache was stored in the latest build, we'll see if it gets restored in the next one

PR #627, opened 2020-06-30
chore(ci): fix node cache

aslafy-z:
Do not install modules if cache hits


.github/workflows/main.yml
```diff
@@ -26,6 +27,7 @@ jobs:
       with:
         node-version: ${{ steps.nvm.outputs.nvmrc }}
     - run: npm ci
+      if: steps.cache-node.outputs.cache-hit != 'true'
```
vikaspotluri123, line 30: Does it make that much of a difference running `npm ci` if there's a cache hit?

aslafy-z: Not a big difference, but still it download all manifests and rebuilds binaries when needed.

PR #628, opened 2020-07-02
Updated data.json by removing deque swag.

Akchy:
The twitter link given is not live and the tweet has been removed

### DIFF
--- a/data.json
+++ b/data.json
@@ -126,15 +126,6 @@
     "tags": ["stickers", "clothing", "hacktoberfest", "expired"]
   },
   {
-    "name": "Deque",
-    "difficulty": "easy",
-    "description": "Answer Deque's feedback survey once you're logged in to get an awesome t-shirt.",
-    "reference": "https://twitter.com/dequesystems/status/1182651282869362694",
-    "image": "https://i.imgur.com/t82DlcW.jpg",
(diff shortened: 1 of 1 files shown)


data.json
```diff
@@ -125,15 +125,6 @@
...
     "tags": ["stickers", "clothing", "hacktoberfest", "expired"]
   },
-  {
-    "name": "Deque",
-    "difficulty": "easy",
-    "description": "Answer Deque's feedback survey once you're logged in to get an awesome t-shirt.",
-    "reference": "https://twitter.com/dequesystems/status/1182651282869362694",
-    "image": "https://i.imgur.com/t82DlcW.jpg",
-    "dateAdded": "2020-02-25T14:33:22.000Z",
-    "tags": ["clothing"]
```
vikaspotluri123, line 135: We normally add the `expired` tag rather than removing an opportunity so people can still reference it if they so choose

// @aslafy-z please correct me if it's different when the reference disappears

aslafy-z: I guess we can keep it. Expired opportunities are skipped in the ci link checking step right?

vikaspotluri123: I don't think so 😬

Akchy: If you wish to see the opportunity for future it's ok, but the it won't be visible anywhere in the Deque social media. If so, you can close this PR

PR #652, opened 2020-08-22
clean up hero and footer

swapagarwal:
- remove github button from hero
- remove logo from footer
- reorder items in footer

Issue #578, opened 2020-05-02
Automatic images upload to imgur


aslafy-z:
- [x] I've checked that this isn't a duplicate issue.

In order to prevent us from broken image links, we could build an automation that upload images to imgur.

vikaspotluri123:
I think we should just cache the images either in Github Actions or Netlify

THE-VR7:
Hello. I am new  and would love to contribute.
I would like to fix this issue, if no one is already on it.

tadhglewis:
As a side note it is against Imgur Terms of Service to use them as a CDN

Stuff not to do
If someone else might own the copyright to it, don't upload it. Don't upload gore, "hate speech" (i.e. demeaning race, gender, age, religious or sexual orientation, etc.), or material that is threatening, harassing, defamatory, or that encourages violence or crime. Don't upload illegal content such as child porn or nonconsensual ("revenge") porn. Don't hotlink to adult content or to file-sharing, gambling, torrent, warez, or Imgur rip-off sites. Don't impersonate someone else. **Also, don't use Imgur to host image libraries you link to from elsewhere, content for your website, advertising, avatars, or anything else that turns us into your content delivery network. If you do – and we will be the judge – or if you do anything illegal, in addition to any other legal rights we may have, we will ban you along with the site you're hotlinking from, delete all your images, report you to the authorities if necessary, and prevent you from viewing any images hosted on Imgur.com. We mean it.**

https://imgur.com/tos

Catalyst497:
Is this issue still open? I would like to work on it. 
I'm a first timer. 
Could you put me through on the details so I could know what to do?

adarshanand67:
Is this issue still open? I would like to work on it.
I'm a first timer.
Could you put me through on the details so I could know what to do?

varunseth5788:
i would like to contribute to this issue.
or need any other help.

Fazal-Abbas2052:
I would like to contribute to this issue. Can you brief me about the details?

NahuelBarriga:
is it still open? I'd love to help

aslafy-z:
This issue is still open. In our data.json, we link to external images. Sometimes, the links breaks. We would like to preserve the images even after they are removed from the original source.

@tadhglewis mentioned that using imgur as a CDN is against their TOS.

It might be better to find another way. Options I see is:
- Automatic issue open on dead link.
- Caching proxy:
[...]

PR #654, opened 2020-08-23
✨ ignore failed images for non-production Netlify deploys

vikaspotluri123:
no issue

We use Netlify to preview changes, which are especially useful in design PRs (e.g. #652). Unfortunately, we depend on third parties for images, and you can never trust the network. This PR drops `gulp-download-stream` in favor of an async, stream-based image downloader. This allows us finer control of the image download pipeline (so in the future, we can also cache images 🥳), and lets the Netlify build pass in non-production environments even if some images are broken

### DIFF
--- a/download-images.js
+++ b/download-images.js
@@ -1,0 +1,61 @@
+const {promisify} = require('util');
+const {pipeline: pipeline_} = require('stream');
+const {createWriteStream} = require('fs');
+const path = require('path');
+const {performance} = require('perf_hooks');
+const Queue = require('p-queue').default;
+const chalk = require('chalk');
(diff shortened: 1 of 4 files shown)


swapagarwal:
> Unfortunately, we depend on third parties for images, and you can never trust the network.

This has hurt us multiple times. Maybe we can move all images to Imgur?

vikaspotluri123:
That was suggested in #578 but it seems like that's against their ToS. I know both Github Actions and Netlify support caching so that's the route I was thinking of going

aslafy-z (approved):

PR #686, opened 2020-09-30
improve docs

parisha99:

Piques could be a tough word for many people so stimulate is a good option.


vikaspotluri123:
@aslafy-z IMO this PR is not in the spirit of hacktoberfest, thoughts?

aslafy-z:
Right. Let's close this PR, sorry @parisha99 but this is not a meaningful PR. There is 30+ issues to solve in this repo, feel free to take one if you want to.

PR #688, opened 2020-09-30
Update Gitpod link

rmorabia:
I was trying to set up development to add new swag opportunities, and found that the Gitpod links aren't working. They redirect to a 404. This is the working format of how Gitpod URLs work. I tested this on my machine and was able to make this PR from Gitpod.


aslafy-z:
This comment raised the issue, well spotted, thanks!
https://github.com/gitpod-io/gitpod/issues/688#issuecomment-691628472

PR #696, opened 2020-10-01
chore(deps): bump gulp-postcss and autoprefixer

vikaspotluri123:
refs #666, #671

PostCSS released a major version which is incompatible with older versions of plugins. We have to bump the plugins together, and both gulp-postcss and autoprefixer now depend on the same major version of postcss


aslafy-z:
Perfect 👍

PR #715, opened 2020-10-01
feat: sort tags

aslafy-z:


vikaspotluri123:
I think at one point we had a requirement for tags to be in alphabetical order but I guess that didn't pan out 😆

PR #737, opened 2020-10-01
fix: set Blockstack opportunity as expired

deadpyxel:

I just checked if I was able to claim the BlockStack stickers opportunity, it seems that web archive link is not working any more. I dared to remove it, but if there's something else that cen be used for that we can update the link.

### DIFF
--- a/data.json
+++ b/data.json
@@ -81,6 +81,15 @@
     "tags": ["clothing", "expired"]
   },
   {
+    "name": "Blockstack",
+    "difficulty": "easy",
+    "description": "Get up to 5 awesome stickers from Blockstack! Don't be evil!",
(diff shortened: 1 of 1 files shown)


aslafy-z:
Hello, we keep history of the changes, please restore the entry and add the `expired` tag instead. Thank you!

vikaspotluri123 (approved):

data.json
```diff
@@ -87,7 +87,7 @@
     "reference": "https://web.archive.org/web/20200616131621/https://blockstack.myshopify.com/collections/frontpage/products/dont-be-evil-stickers",
     "image": "https://i.imgur.com/sHgpibu.jpg",
     "dateAdded": "2020-07-21T14:42:00.000Z",
-    "tags": ["stickers"]
+    "tags": ["stickers, expired"]
```
aslafy-z, line 90:
```suggestion
    "tags": ["stickers", "expired"]
```

PR #785, opened 2020-10-20
chore: add test for image mime type

### LINKED ISSUE #761: Ensure mime type of image resources
Add test that checks if the image field is an accepted image format: https://sharp.pixelplumbing.com/#formats

ritesh-pandey:

Fixes #761 
Added test for image mime type.

### DIFF
--- a/test/functional/data.spec.js
+++ b/test/functional/data.spec.js
@@ -61,13 +61,7 @@ describe('swag-for-dev', function () {
 						});
 					}
 
-					it(opportunity.name + ' has a valid image URL', function () {
-						this.timeout(10000);
-						this.slow(1500);
-						return checkURL(opportunity.image);
-					});
-
-					it(opportunity.name + ' has a valid image format', function (done) {
+					it(opportunity.name + ' has a valid image', function (done) {
 						this.timeout(10000);
 						this.slow(1500);
 						got(opportunity.image)
--- a/test/functional/data.spec.js
+++ b/test/functional/data.spec.js
@@ -61,23 +61,12 @@ describe('swag-for-dev', function () {
 						});
 					}
 
-					it(opportunity.name + ' has a valid image', function (done) {
+					it(opportunity.name + ' has a valid image', async function () {
 						this.timeout(10000);
 						this.slow(1500);
-						got(opportunity.image)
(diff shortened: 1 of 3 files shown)


test/functional/data.spec.js
```diff
@@ -60,11 +61,30 @@ describe('swag-for-dev', function () {
 						});
 					}
 
-					it(opportunity.name + ' has a valid image', function () {
+					it(opportunity.name + ' has a valid image URL', function () {
```
vikaspotluri123, line 64: From my perspective, there's no need to check if the image URL is valid since we're checking the mime type of the image - by checking the mime type of the image, we're implicitly validating that the image also exists

ritesh-pandey: I added new test for mime-type without modifying existing ones. Just changed test description text.

I agree to your point. Downloading and checking an image means implicitly checking for valid URL.

test/functional/data.spec.js
```diff
@@ -60,10 +61,23 @@ describe('swag-for-dev', function () {
...
+									.then(function () {
+										done();
+									})
+									.catch(function (err) {
+										done(err);
+									});
+							})
+							.catch(function (err) {
+								done(err);
+							});
```
vikaspotluri123, line 80:
```suggestion
						const {rawBody} = await got(opportunity.image)
						await sharp(response.rawBody, {limitInputPixels: false}).toBuffer()
```

test/functional/data.spec.js
```diff
@@ -60,10 +61,23 @@ describe('swag-for-dev', function () {
 						});
 					}
 
-					it(opportunity.name + ' has a valid image', function () {
+					it(opportunity.name + ' has a valid image', function (done) {
```
vikaspotluri123, line 64:
```suggestion
					it(opportunity.name + ' has a valid image', async function () {
```

test/functional/data.spec.js
ritesh-pandey: To maintain homogeneity with previous code, I used Promise instead of async/await.

vikaspotluri123 (approved):

vikaspotluri123:
@aslafy-z merge when you get a chance!

PR #797, opened 2020-11-06
feat: add JSON schema validation

ritesh-pandey:

Added AJV-based JSON schema validation.


package.json
```diff
@@ -14,6 +14,7 @@
     "url": "git+https://github.com/swapagarwal/swag-for-dev.git"
   },
   "dependencies": {
+    "ajv": "^6.12.6",
```
aslafy-z, line 17: This should go in `devDependencies` instead

schema.json
```diff
@@ -0,0 +1,107 @@
...
+                        "items": {
+                            "$id": "#/items/anyOf/0/properties/tags/items",
+                            "anyOf": [
+                                {
+                                    "$id": "#/items/anyOf/0/properties/tags/items/anyOf/0",
+                                    "type": "string",
+                                    "title": "The first anyOf schema",
+                                    "description": "An explanation about the purpose of this instance.",
+                                    "default": "",
+                                    "examples": []
```
aslafy-z, line 98: I'm wondering if we should not set a fixed list of allowed tags in there. @vikaspotluri123 wdyt?

vikaspotluri123: It's not a fixed list, is it?

test/functional/data.spec.js
```diff
@@ -23,6 +25,12 @@ function checkURL(url, head = false) {
 }
 
 describe('swag-for-dev', function () {
+	it('data.json has valid schema', function () {
+		const ajv = new Ajv();
+		const valid = ajv.validate(schema, data);
+		expect(valid).to.be.true;
+	});
+
 	it('data.json is valid', function () {
```
aslafy-z, line 34: This test will be replaced by 'data.json has valid schema'.

vikaspotluri123: I think this should be merged with the schema test since it checks that everything is in alphabetical order

test/functional/data.spec.js
```diff
@@ -2,8 +2,10 @@ const got = require('got');
 const {expect} = require('chai');
 const parallel = require('mocha.parallel');
 const sharp = require('sharp');
+const Ajv = require('ajv');
 
 const data = require('../../data.json');
```
aslafy-z, line 7: I'd say this require should go into a test, so tests wont break if json is semantically not valid

ritesh-pandey: Didn't get this point. Why would test break with a require statement?

test/functional/data.spec.js
ritesh-pandey: `chai-json-schema-ajv` plug-in should be a good fit for running tests.

vikaspotluri123: No need to add an extra dep 😄 what you have now for validation is good / on the right track 👍🏾

test/functional/data.spec.js
vikaspotluri123: @aslafy-z I think we decided to require it outside because if the json is not valid the test should fail?

aslafy-z (approved):

test/functional/data.spec.js
```diff
@@ -23,6 +25,12 @@ function checkURL(url, head = false) {
 }
 
 describe('swag-for-dev', function () {
+	it('data.json has valid schema', function () {
+		const ajv = new Ajv();
```
aslafy-z, line 29:
```suggestion
		const ajv = new Ajv(); 
```

test/functional/data.spec.js
```diff
@@ -23,6 +25,12 @@ function checkURL(url, head = false) {
...
```
aslafy-z, line 29:
```suggestion
		const ajv = new Ajv();
```

schema.json
```diff
@@ -0,0 +1,108 @@
...
+                        "default": "",
+                        "examples": []
+                    },
+                    "reference": {
+                        "$id": "#/items/anyOf/0/properties/reference",
+                        "type": "string",
+                        "title": "The reference schema",
+                        "description": "An explanation about the purpose of this instance.",
+                        "default": "",
+                        "examples": []
```
aslafy-z, line 61:
```suggestion
                        "examples": [],
                        "format": "uri"
```

aslafy-z:
@ritesh-pandey Can you please rebase your branch so we can merge your PR? 
Thank you for the awesome contribution!

ritesh-pandey:
Rebased the branch. The error message needs improvement. It is not very intuitive at the moment.

anandfresh (approved):

Issue #881, opened 2021-10-03
Blitz.js free stickers


martinsaxa:
## Wild swag opportunity appeared!

- [x ] I've checked that this isn't a duplicate issue.

Q. Which company offers it?  
A. : Prisma

Q. Where have you seen the announcement (URL)?  
A. : https://blitzjs.com/stickers

Q. What kind of swag (clothing, stickers, ...)?  
A. : stickers

Q. Have you received swag from this?  
A. : no

Q. How long did it take?  
A. : n/a

Q. Does it ship worldwide?  
A. : yes

- [ x] I've checked that the promotion is legit.
- [ x] I've checked that it's still available to the public.

hyperhtml:
Added blitz.js stickers a PR #882

aslafy-z:
This is now expired. Thank you for your participation.

PR #882, opened 2021-10-03
feat: Added blitz.js stickers

hyperhtml:

Added blitz.js stickers, noted in #881, and referenced initial twitter post regarding the sticker offering.


aslafy-z:
This is now expired. Thank you for your participation.

PR #1062, opened 2023-07-11
Update data.spec.js

GAUTAMMANAK1:
Removed unnecessary line breaks and adjusted the code formatting for better readability. Adjusted the loop condition in the for loop to use i < data.length instead of i < length for consistency. Removed the eslint-disable and eslint-enable comments as they are not needed. Removed the unnecessary comments since the code is self-explanatory. Added appropriate spacing to enhance code readability. With these changes, the code will have the same functionality and work as before, but with improved code formatting and readability.

### DIFF
--- a/test/functional/data.spec.js
+++ b/test/functional/data.spec.js
@@ -1,5 +1,5 @@
 const got = require('got');
-const {expect} = require('chai');
+const { expect } = require('chai');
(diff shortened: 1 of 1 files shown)


GAUTAMMANAK1:
To modify the code while keeping the same functionality and work

PR #1064, opened 2023-10-01
fix: checkbox bg color white issue

### LINKED ISSUE #1063: [Bug] : checkbox background is not fully blue when we select done option.
## Issue regarding the website

- [Y] I've checked that this isn't a duplicate issue.

When we select in done checkbox there is a small white dot left it should be blue as background.

## screenshot ##

white

i have highlited the error you can see clearly here there is a small white dot left when user checked the boxes the issue remains in all checkbox.

love-sonkar:
fix #1063 checkbox background is not fully blue when we select done option now the background is fully colord as secondary bg in file

- [Y] I've checked that this isn't a duplicate pull request.

i have chenged card.styl file to fix this issue

### DIFF
--- a/src/styl/imports/card.styl
+++ b/src/styl/imports/card.styl
@@ -172,6 +172,7 @@
 				border .8em solid secondary-color
 				animation bounce 200ms cubic-bezier(.4,.0,.23,1)
 				opacity 1
+				background-color secondary-color				
 
 				&::before
 					content ''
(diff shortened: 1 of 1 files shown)


src/styl/imports/card.styl
```diff
@@ -172,6 +172,7 @@
 				border .8em solid secondary-color
 				animation bounce 200ms cubic-bezier(.4,.0,.23,1)
 				opacity 1
+				background-color secondary-color
```
aslafy-z, line 175:
```suggestion
				background-color secondary-color
```

aslafy-z (changes requested):

aslafy-z (approved):
LGTM, thank you!

PR #1066, opened 2023-10-05
Added swag opportunity[Amplication]

utsavbhattarai007:

I've updated the `data.json` file by adding the swag opportunity from Amplication.

### DIFF
--- a/data.json
+++ b/data.json
@@ -36,6 +36,15 @@
     "tags": ["stickers"]
   },
   {
+    "name": "Amplication",
+    "difficulty": "medium",
+    "description": "Participate in the Amplication hacktoberfest challenge! Propose an app idea, get it approved, and build it using Amplication. Then, craft a clear and comprehensive tutorial or video walkthrough. Post it on your preferred platform and submit a PR link to the official repo. Get a chance to have your tutorial featured and receive a T-shirt for your effort!",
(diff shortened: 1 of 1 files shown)


data.json
```diff
@@ -35,6 +35,15 @@
     "dateAdded": "2018-12-09T07:08:00.000Z",
     "tags": ["stickers"]
   },
+  {
+    "name": "Amplication",
+    "difficulty": "medium",
+    "description": "Participate in the Amplication hacktoberfest challenge! Propose an app idea, get it approved, and build it using Amplication. Then, craft a clear and comprehensive tutorial or video walkthrough. Post it on your preferred platform and submit a PR link to the official repo. Get a chance to have your tutorial featured and receive a T-shirt for your effort!",
+    "reference": "https://github.com/amplication/amplication/issues/7028",
```
aslafy-z, line 42:
```suggestion
    "reference": "https://github.com/amplication/amplication/issues/7026",
```

aslafy-z: It looks like this is the correct link

aslafy-z (changes requested):

data.json
utsavbhattarai007: No, For the description as I mentioned above, the above link is right. The link you have mentioned is for winning the swag by solving the issue not by building the app

aslafy-z (approved):

PR #1193, opened 2023-10-22
Update data.json

RS-labhub:

Added "Coding ninjas"


aslafy-z:
Thanks for your submission, 🙏, but we can't list opportunities tied to contests or raffles with exclusive eligibility requirements. We want to keep the experience open and enjoyable for everyone. Unfortunately, we can't proceed with this one. 

If you think this opportunity should be part of the list, please open an issue so we can discuss about it first. 
Thank you

Issue #1220, opened 2023-10-24
[image] In Novu the image is not fitting in frame.


BandhiyaHardik:
## Issue with an existing swag opportunity

- [x] I've checked that this isn't a duplicate issue.

Q. Which opportunity?  
A. : Novu

There is no issue with swag opportunity.

i was looking into the website and i found in many swags the images are not fitting in the frame.

Snapshot of image on website:
![image](https://github.com/swapagarwal/swag-for-dev/assets/110784317/8bb5e57a-00a1-4c2a-88ee-8471a2f5df4b)

The Actual Image:
![image](https://github.com/swapagarwal/swag-for-dev/assets/110784317/64d7dd1b-0332-41cf-a114-d0696ea3565f)

## Solution:
By resizing the image.

BandhiyaHardik:
i would like to work on this issue

PR #1238, opened 2023-10-27
fix: [image] In Novu the image is not fitting in frame.

BandhiyaHardik:
In Novu the image is not fitting in frame.

This fixes https://github.com/swapagarwal/swag-for-dev/issues/1220


BandhiyaHardik:
@aslafy-z can u please look into this error ?

aslafy-z:
@BandhiyaHardik Looks like it was a transient error. It passes now!

BandhiyaHardik:
@aslafy-z  Thank You for always helping !!

Issue #1268, opened 2024-10-02
gatsby swag is no more avaliable


Tejas-Sharma-dev:
## Issue with an existing swag opportunity

- [ ] I've checked that this isn't a duplicate issue.

Q. Which opportunity?  
A. : gatsby swag

![image](https://github.com/user-attachments/assets/0b3a3bf3-505e-4540-be7c-7e480662aa1d)

The gatsby cloud, which offered various swags to contributor, is no more (they are shutdown and acquired by netlify).
the swag is expired.

![image](https://github.com/user-attachments/assets/52476b70-d6da-46c1-b9b7-02acbba66125)

aslafy-z:
I would be pleased to review a PR that sets this opportunity as expired. Let me know if any extra guidance is needed.

Tejas-Sharma-dev:
Hi @aslafy-z , 

the Gatsby cloud swag is avaliable again, they have provided new links to avail swags. shall i create new issue, or submit direct PR, with updated links?

updated-link: https://www.gatsbyjs.com/contributing/contributor-swag/

aslafy-z:
@Tejas-Sharma-dev nice catch!
You can open a pr directly

Tejas-Sharma-dev:
cool, I'll work on it :)

PR #1272, opened 2024-10-04
fix: gatsby swag revival

Tejas-Sharma-dev:

updated data.json for 'gatsby' swag, which was marked as 'expired' (ref: #1268)

- updated the swag link
- removed 'expired' tag

### DIFF
--- a/data.json
+++ b/data.json
@@ -427,9 +427,9 @@
     "difficulty": "hard",
     "description": "Get a pull request merged into Gatsby and receive a free shirt, socks, or stickers.",
     "reference": "https://www.gatsbyjs.com/contributing/contributor-swag/",
-    "image": "https://i.imgur.com/e7GwaTN.png",
+    "image": "https://cdn.shopify.com/s/files/1/0332/2730/9195/products/Shirt.png?v=1677684481&width=600&height=",
     "dateAdded": "2024-10-04T16:08:09.000Z",
-    "tags": ["clothing", "stickers"]
+    "tags": ["clothing", "stickers", "mug"]
   },
   {
     "name": "GeeksForGeeks",
--- a/data.json
+++ b/data.json
@@ -426,10 +426,10 @@
     "name": "Gatsby",
     "difficulty": "hard",
     "description": "Get a pull request merged into Gatsby and receive a free shirt, socks, or stickers.",
-    "reference": "https://www.gatsbyjs.org/docs/contributor-swag/",
+    "reference": "https://www.gatsbyjs.com/contributing/contributor-swag/",
     "image": "https://i.imgur.com/e7GwaTN.png",
     "dateAdded": "2018-10-24T11:42:09.000Z",
-    "tags": ["clothing", "stickers", "expired"]
+    "tags": ["clothing", "stickers"]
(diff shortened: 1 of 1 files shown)


data.json
```diff
@@ -426,10 +426,10 @@
     "name": "Gatsby",
     "difficulty": "hard",
     "description": "Get a pull request merged into Gatsby and receive a free shirt, socks, or stickers.",
-    "reference": "https://www.gatsbyjs.org/docs/contributor-swag/",
+    "reference": "https://www.gatsbyjs.com/contributing/contributor-swag/",
     "image": "https://i.imgur.com/e7GwaTN.png",
     "dateAdded": "2018-10-24T11:42:09.000Z",
```
aslafy-z, line 431: Please update dateAdded to now please. We might want to rename or add another field for the updated date in the future.

data.json
```diff
@@ -426,10 +426,10 @@
     "name": "Gatsby",
     "difficulty": "hard",
     "description": "Get a pull request merged into Gatsby and receive a free shirt, socks, or stickers.",
-    "reference": "https://www.gatsbyjs.org/docs/contributor-swag/",
+    "reference": "https://www.gatsbyjs.com/contributing/contributor-swag/",
```
aslafy-z, line 429: It looks like the actual swag is Netlify swag, not Gatsby one. I think we want to update the Netlify entry instead and keep that one expired.

Tejas-Sharma-dev: sounds good, why don't we update as 'Gatsby x Netlify' swag?

aslafy-z: It's up to you

Tejas-Sharma-dev: fine then, I'II just update the name and date. instead of adding new swag.

data.json
```diff
@@ -423,13 +423,13 @@
     "tags": ["clothing", "hacktoberfest", "stickers", "accessories", "expired"]
   },
   {
-    "name": "Gatsby",
+    "name": "Gatsby x Netlify",
     "difficulty": "hard",
     "description": "Get a pull request merged into Gatsby and receive a free shirt, socks, or stickers.",
-    "reference": "https://www.gatsbyjs.org/docs/contributor-swag/",
+    "reference": "https://www.gatsbyjs.com/contributing/contributor-swag/",
     "image": "https://i.imgur.com/e7GwaTN.png",
```
aslafy-z, line 430: Can you please also update the image with one from the store to match the new swags

data.json
```diff
@@ -423,13 +423,13 @@
...
-    "dateAdded": "2018-10-24T11:42:09.000Z",
-    "tags": ["clothing", "stickers", "expired"]
+    "dateAdded": "2024-10-04T16:08:09.000Z",
+    "tags": ["clothing", "stickers"]
```
aslafy-z, line 432:
```suggestion
    "tags": ["clothing", "mug", "stickers"]
```

data.json
Tejas-Sharma-dev: ok, i'm updating image URL

aslafy-z:
Thank you @Tejas-Sharma-dev!

